PR #17, opened 2021-09-28
Address docs build warnings

bwohlberg:
Address docs build warnings due to `NEP35` cross reference in docstrings of some functions imported from `numpy`.


bwohlberg:
Included changes from #16 by mistake.

PR #25, opened 2021-10-05
Allow ADMM f to be a Functional

bwohlberg:
Change typing to allow f to be any `Functional` rather than requiring it to be a `Loss`.


lukepfister:
LGTM, but this needs to be rebased onto main before it can be merged

Issue #38, opened 2021-10-11
Scale problem in ADMM / CG


tbalke:
@bwohlberg @lukepfister @Michael-T-McCann 

Scaling the f-loss seems to scale the reciprocal of the reconstruction (see below). Initially, I brought this up when designing the `svmbir` PPP example but now I found out that it is not limited to that. It seems like the problem should be somewhere in the CG but not sure.

Lastly, using `subproblem_solver=None` produces `nan` as a result which may or may not be from the same origin.

```python
import jax
import numpy as np
from scico import functional, linop, loss
from scico.admm import ADMM, LinearSubproblemSolver

np.random.seed(123)
N = 10
M = 100

x = jax.device_put(np.random.randn(N))
A = linop.MatrixOperator(np.random.randn(M,N))
y = A @ x

f1 = loss.SquaredL2Loss(y=y, A=A, scale=1)
f42 = loss.SquaredL2Loss(y=y, A=A, scale=42)

x1 = ADMM(
    f=f1,
    g_list=[functional.ZeroFunctional()],
    C_list=[linop.Identity(x.shape)],
    rho_list=[1],
    subproblem_solver=LinearSubproblemSolver(cg_function='scico'), #same with 'jax'
).solve()

x42 = ADMM(
    f=f42,
    g_list=[functional.ZeroFunctional()],
    C_list=[linop.Identity(x.shape)],
    rho_list=[1],
    subproblem_solver=LinearSubproblemSolver(cg_function='scico'), #same with 'jax'
).solve()

print(np.mean(x1/x42))

# 41.99956
```

tbalke:
@lukepfister Yes, that did it!! Will create a PR.

bwohlberg:
Resolved by #45

PR #45, opened 2021-10-14
Fix scaling bug in CG

tbalke:
@bwohlberg @lukepfister 
Kept the `scale` factor in the Hessian the same so that it truly is still the Hessian but removed the other `scale` factor.


bwohlberg:
Resolves #38

bwohlberg:
Can we add a test that would catch the bug?

bwohlberg:
> Change the loss scale to something like .3333 in the existing ADMM quadratic tests and that should do the trick

That doesn't do the trick, unfortunately. See branch `brendt/admm-quadratic-test`, which was made off the branch for this PR. The modified tests pass for `GenericSubproblemSolver` but they fail for `LinearSubproblemSolver`, so either (a) this PR doesn't resolve the problem, or (b) there's something wrong with the tests. If the tests and fix are correct, the tests involving `GenericSubproblemSolver` should fail for the `main` branch but pass in the branch for this PR.

bwohlberg:
> No, `brendt/admm-quadratic-test` branches of a different commit.

You're right, I branched it off the other `tbalke/...` branch by mistake.
 
> When using the fix from this PR most of the tests from `brendt/admm-quadratic-test` pass again, including the one with `LinearSubproblemSolver`. (Not all tests pass though)

All tests pass for me now that I recreated my branch as a branch off of `tbalke/cgfix`.

> scico/test/test_admm.py::TestComplex::test_admm_generic FAILED                                                                                

Could you check the detailed log on this test? Perhaps it's failing because the accuracy tolerance is a bit too tight?

tbalke:
> All tests pass for me now that I recreated my branch as a branch off of `tbalke/cgfix`.

For me, too, all tests pass now. Thanks!

PR #46, opened 2021-10-16
Improve ADMM tests

bwohlberg:
Modify ADMM tests to include tests for `LinearSubproblemSolver` with `SquaredL2Loss` with a non-default scaling factor.


bwohlberg:
See also #45 and #38.

PR #74, opened 2021-11-02
Microscopy example scripts

bwohlberg:
- Add a new microscopy example scripts that uses all three data channels
- Minor improvements to existing single-channel example script


Michael-T-McCann (approved):
The result definitely looks more impressive in color. I fixed what looked like a typo.

PR #77, opened 2021-11-08
A collection of small docs updates

Michael-T-McCann:
Typos, clarifications, etc. Placed here to avoid cluttering the history with small commits.


scico/solver.py
```diff
@@ -257,20 +257,20 @@ def cg(
 ) -> Union[JaxArray, dict]:
     r"""Conjugate Gradient solver.
 
-    Solve the linear system :math:`A\mb{x} = \mb{b}` via the conjugate
-    gradient method.
+    Solve the linear system :math:`A\mb{x} = \mb{b}`, where :math:`A` is
+    positive definite, via the conjugate gradient method.
```
lukepfister, line 261: needs to be symmetric, too

Michael-T-McCann: Confusingly, I think most authors make positive definite operators a subset of self-adjoint operators, so positive definite -> symmetric.

PR #78, opened 2021-11-09
Problem with W^1/2 weight exponent

tbalke:
Initially I thought `svmbir` and `scico` use different notation for weighted norms, so I ignored this. Now after second thought I think we need remove the squaring of the weights.

Our weight_op is `||y||_W^2 = ||W^(1/2) y||^2 = y^T W y` so there is no need to square the weights before passing them to svmbir. (Albeit in the svmbir docs the notation was not really clear until I updated it just now.)

In the example scripts I removed the square root operation on the weights to compensate for the squaring. This, however, yields different regularization for the cg-example, so it requires retuning. I suggest to return to this later, since I am expecting there may be more changes to the svmbir interface in the near future which would require another retuning.

### DIFF
--- a/examples/scripts/ct_astra_weighted_tv_admm.py
+++ b/examples/scripts/ct_astra_weighted_tv_admm.py
@@ -137,7 +137,7 @@ where
 """
 lambda_weighted = 1.14e2
 
-weights = counts / Io  # scale by Io to balance the data vs regularization term
+weights = counts / Io  # scaled by Io to balance the data vs regularization term
 f = loss.WeightedSquaredL2Loss(y=y, A=A, W=linop.Diagonal(weights))
 
 admm_weighted = ADMM(
--- a/examples/scripts/ct_astra_weighted_tv_admm.py
+++ b/examples/scripts/ct_astra_weighted_tv_admm.py
@@ -138,8 +138,7 @@ where
 lambda_weighted = 1.14e2
 
 weights = counts / Io  # scale by Io to balance the data vs regularization term
-W = linop.Diagonal(snp.sqrt(weights))
-f = loss.WeightedSquaredL2Loss(y=y, A=A, W=W)
+f = loss.WeightedSquaredL2Loss(y=y, A=A, W=linop.Diagonal(weights))
 
 admm_weighted = ADMM(
     f=f,
--- a/examples/scripts/ct_astra_weighted_tv_admm.py
+++ b/examples/scripts/ct_astra_weighted_tv_admm.py
@@ -139,7 +139,7 @@ lambda_weighted = 1.14e2
 
 weights = counts / Io  # scale by Io to balance the data vs regularization term
 W = linop.Diagonal(snp.sqrt(weights))
-f = loss.WeightedSquaredL2Loss(y=y, A=A, weight_op=W)
+f = loss.WeightedSquaredL2Loss(y=y, A=A, W=W)
 
 admm_weighted = ADMM(
     f=f,
--- a/scico/loss.py
+++ b/scico/loss.py
@@ -211,7 +211,7 @@ class WeightedSquaredL2Loss(Loss):
         if W is None:
             self.W = linop.Identity(y.shape)
         elif isinstance(W, linop.Diagonal):
-            if np.all(W.diagonal >= 0):
+            if snp.all(W.diagonal >= 0):
                 self.W = W
             else:
                 raise Exception(f"The weights, W.diagonal, must be non-negative.")
--- a/scico/loss.py
+++ b/scico/loss.py
@@ -55,7 +55,7 @@ class Loss(functional.Functional):
         r"""Initialize a :class:`Loss` object.
 
         Args:
-            y : Measurements
+            y : Measurement.
             A : Forward operator.  Defaults to None.  If None, ``self.A`` is a :class:`.Identity`.
             scale : Scaling parameter.  Default: 0.5.
 
@@ -122,7 +122,7 @@ class SquaredL2Loss(Loss):
         r"""Initialize a :class:`SquaredL2Loss` object.
 
         Args:
-            y : Measurements.
+            y : Measurement.
             A : Forward operator.  If None, defaults to :class:`.Identity`.
             scale : Scaling parameter.
         """
@@ -197,7 +197,7 @@ class WeightedSquaredL2Loss(Loss):
         r"""Initialize a :class:`WeightedSquaredL2Loss` object.
 
         Args:
-            y : Measurements.
+            y : Measurement.
             A : Forward operator.  If None, defaults to :class:`.Identity`.
             scale : Scaling parameter.
             W:  Weighting diagonal operator. Must be non-negative.
@@ -286,7 +286,7 @@ class PoissonLoss(Loss):
         r"""Initialize a :class:`Loss` object.
 
         Args:
-            y : Measurements
+            y : Measurement.
             A : Forward operator.  Defaults to None.  If None, ``self.A`` is a :class:`.Identity`.
             scale : Scaling parameter.  Default: 0.5.
 
--- a/scico/loss.py
+++ b/scico/loss.py
@@ -272,7 +272,6 @@ class PoissonLoss(Loss):
 
     .. math::
         \mathrm{scale} \sum_i [A(x)]_i - y_i \log\left( [A(x)]_i \right) + \log(y_i!)
-
     """
 
     def __init__(
@@ -285,15 +284,22 @@ class PoissonLoss(Loss):
 
         Args:
             y : Measurement.
-            A : Forward operator.  Defaults to None.  If None, ``self.A`` is a :class:`.Identity`.
-            scale : Scaling parameter.  Default: 0.5.
-
+            A : Forward operator. Defaults to None.  If None, ``self.A`` is a :class:`.Identity`.
+            scale : Scaling parameter. Default: 0.5.
         """
         y = ensure_on_device(y)
         super().__init__(y=y, A=A, scale=scale)
 
         #: Constant term in Poisson log likehood; equal to ln(y!)
         self.const: float = gammaln(self.y + 1)  # ln(y!)
+
+        # The "true" Poisson loss is not differentiable at zero, but the
+        # ε that we add to allow evaluation at zero has the side-effect
+        # of making it differentiable there.
+        if isinstance(A, operator.Operator):
+            self.is_smooth = A.is_smooth
+        else:
+            self.is_smooth = None
 
     def __call__(self, x: Union[JaxArray, BlockArray]) -> float:
         ε = 1e-9  # So that loss < infinity
--- a/scico/loss.py
+++ b/scico/loss.py
@@ -127,7 +127,6 @@ class SquaredL2Loss(Loss):
             scale : Scaling parameter.
         """
         y = ensure_on_device(y)
-        self.functional = functional.SquaredL2Norm()
         super().__init__(y=y, A=A, scale=scale)
 
         if isinstance(A, operator.Operator):
@@ -207,7 +206,6 @@ class WeightedSquaredL2Loss(Loss):
 
         self.W: linop.Diagonal
 
-        self.functional = functional.SquaredL2Norm()
         if W is None:
             self.W = linop.Identity(y.shape)
         elif isinstance(W, linop.Diagonal):
--- a/scico/loss.py
+++ b/scico/loss.py
@@ -119,6 +119,13 @@ class SquaredL2Loss(Loss):
         A: Optional[Union[Callable, operator.Operator]] = None,
         scale: float = 0.5,
     ):
+        r"""Initialize a :class:`SquaredL2Loss` object.
+
+        Args:
+            y : Measurements.
+            A : Forward operator.  If None, defaults to :class:`.Identity`.
+            scale : Scaling parameter.
+        """
         y = ensure_on_device(y)
         self.functional = functional.SquaredL2Norm()
         super().__init__(y=y, A=A, scale=scale)
@@ -140,7 +147,7 @@ class SquaredL2Loss(Loss):
         Args:
             x : Point at which to evaluate loss.
         """
-        return self.scale * self.functional(self.y - self.A(x))
+        return self.scale * (snp.abs(self.y - self.A(x)) ** 2).sum()
 
     def prox(self, x: Union[JaxArray, BlockArray], lam: float) -> Union[JaxArray, BlockArray]:
         if isinstance(self.A, linop.Diagonal):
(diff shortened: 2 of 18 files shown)


tbalke:
It fails the `test_prox_weights` now which may not be a bad thing as we are still thinking that the prox is incorrect due to the rot_radius issue. Opinions?

bwohlberg:
> It fails the `test_prox_weights` now which may not be a bad thing as we are still thinking that the prox is incorrect due to the rot_radius issue. Opinions?

If that were the explanation, then I would expect that the test without weights would also be failing. Perhaps it's simply that the error due to unclear svmbir docs was also propagated into the tests?
@Michael-T-McCann : since you wrote the tests for this (I think), could you please take a look?

Michael-T-McCann:
The relevant documentation is https://svmbir.readthedocs.io/en/latest/theory.html and https://scico.readthedocs.io/en/latest/_autosummary/scico.loss.html#scico.loss.WeightedSquaredL2Loss. What I see is that SVMBIR and SCICO use the same definition for the weighted norm, but in SVMBIR you pass the weight matrix ("$\Lambda$ corresponds to `weights`") and in SCICO you pass the square root of the weight matrix "`weight_op` ... Corresponds to $W^{1/2}".

Hence https://github.com/lanl/scico/blob/385254282dd82cf4b7ef279cebe7d50e9e9be757/scico/linop/radon_svmbir.py#L128-L130 in our wrapper.

tbalke:
Now I am confused. Are we really handling this consistently?

In ```class WeightedSquaredL2Loss(Loss):``` on one hand we have
https://github.com/lanl/scico/blob/385254282dd82cf4b7ef279cebe7d50e9e9be757/scico/loss.py#L227-L236
So the part ```lhs = c * 2.0 * A.conj() * W * W.conj() * self.y + x``` seems to align with @Michael-T-McCann's assessment.
But then we have
https://github.com/lanl/scico/blob/385254282dd82cf4b7ef279cebe7d50e9e9be757/scico/loss.py#L239-L255
so the line ```eval_fn=lambda x: 2 * self.scale * self.A.adj(self.weight_op(self.A(x))),``` seems to instead use the other version.

In other words we use `A^T W W^T y` in one case and `A^T W A x` in the other, which looks inconsistent.

tbalke:
I am starting to believe that my changes in this PR are incorrect, and you were right, @Michael-T-McCann.

(1) Let's start with that the loss is ```a || Ax-y ||^2_W```. Then according to
https://github.com/lanl/scico/blob/385254282dd82cf4b7ef279cebe7d50e9e9be757/scico/loss.py#L224-L225
we have that `weight_op.diagonal = W^(1/2)`.

(2) The prox in
https://github.com/lanl/scico/blob/385254282dd82cf4b7ef279cebe7d50e9e9be757/scico/loss.py#L227-L236
is correct but the choice of variable names is deeply confusing. It would be better to have something like
```python
            sqrtW = self.weight_op.diagonal
            lhs = c * 2.0 * A.conj() * sqrtW * sqrtW.conj() * self.y + x
```

(3) The Hessian seems to be incorrect. With the definition of the loss in (1) the Hessian would be ```2a A^T W A```, which translates to ```2 * scale * A.adj(weight_op(weight_op(A(x))))```. However, currently we have
https://github.com/lanl/scico/blob/385254282dd82cf4b7ef279cebe7d50e9e9be757/scico/loss.py#L249

Michael-T-McCann:
Nice progress on this! A few thoughts.
>     * Why is the weight operation to be kept so general rather than using a matrix operation?
I don't know what you mean by "using a matrix operation." That said, it seems to me we only ever use diagonal weightings. I would assume the class was written to allow nondiagnal weightings simply because it is mathematically defined. To discuss: simplifying this to diagonal only?
>     * If we use a matrix operation, would it make more sense to store `W` directly (rather than `W^(1/2)`?
I suspect storing the square root was as method to enforce positive definiteness. Whether we keep that depends on the first point. If we do keep it, holy cow we need to change the variable name to something like `W_sqroot`. The documentation is also too subtle.
>     * The Hessian also never gets tested according to codecov, which should be changed
To take this a step further: if the Hessian is neither tested nor (apparently) used anywhere, we should consider dropping it. If we decide to keep it, agreed it needs tests.

bwohlberg:
Are there really only two example scripts that are affected by the change in the meaning of `W`?

scico/loss.py
```diff
@@ -140,7 +147,7 @@ def __call__(self, x: Union[JaxArray, BlockArray]) -> float:
         Args:
             x : Point at which to evaluate loss.
         """
-        return self.scale * self.functional(self.y - self.A(x))
+        return self.scale * (snp.abs(self.y - self.A(x)) ** 2).sum()
```
bwohlberg, line 151: Is this change desirable? Dropping the use if `self.functional` seems like it may have consequences for derived classes.

scico/loss.py
```diff
@@ -154,7 +161,7 @@ def prox(self, x: Union[JaxArray, BlockArray], lam: float) -> Union[JaxArray, Bl
     @property
     def hessian(self) -> linop.LinearOperator:
         r"""If ``self.A`` is a :class:`.LinearOperator`, returns a new :class:`.LinearOperator` corresponding
-        to Hessian :math:`\mathrm{A^*A}`.
+        to Hessian :math:`2 \mathrm{scale} \cdot \mathrm{A^* A}`.
```
bwohlberg, line 165: Docs were wrong in omission of 2 * scale? Also, not sure the `\cdot` is notationally necessary for scalar multiplication.

scico/loss.py
```diff
@@ -184,30 +191,33 @@ def __init__(
...
         scale: float = 0.5,
-        weight_op: Optional[operator.Operator] = None,
+        W: Optional[linop.Diagonal] = None,
     ):
 
         r"""Initialize a :class:`WeightedSquaredL2Loss` object.
 
         Args:
-            y : Measurements
+            y : Measurements.
```
bwohlberg, line 200: Nitpick: change to "Measurement."

scico/loss.py
```diff
@@ -218,40 +228,43 @@ def __init__(
         if isinstance(self.A, linop.LinearOperator):
             self.is_quadratic = True
 
-        if isinstance(self.A, linop.Diagonal) and isinstance(self.weight_op, linop.Diagonal):
+        if isinstance(self.A, linop.Diagonal) and isinstance(self.W, linop.Diagonal):
             self.has_prox = True
 
     def __call__(self, x: Union[JaxArray, BlockArray]) -> float:
-        return self.scale * self.functional(self.weight_op(self.y - self.A(x)))
+        return self.scale * (self.W.diagonal * snp.abs(self.y - self.A(x)) ** 2).sum()
```
bwohlberg, line 236: See earlier comment on similar lines.

examples/scripts/ct_astra_weighted_tv_admm.py
```diff
@@ -139,7 +139,7 @@ def postprocess(x):
 
 weights = counts / Io  # scale by Io to balance the data vs regularization term
 W = linop.Diagonal(snp.sqrt(weights))
```
bwohlberg, line 141: Should't the `snp.sqrt` be dropped here?

bwohlberg:
CI seems to be failing because the changes would slightly reduce the test coverage percentage.

scico/loss.py
tbalke: Yes, they were incorrect.

I added the `\cdot` to keep things consistent with other parts of the repo. I agree though, we could remove all of the `\cdot`s where  the are unnecessary.

scico/loss.py
tbalke: fixed in 7db54df

scico/loss.py
tbalke: I don't think it does. Does it?

I have this to keep it consistent with the weighted case where we can avoid computing the square root of the weights.

examples/scripts/ct_astra_weighted_tv_admm.py
tbalke: Good catch! Addressed in 911e846.

scico/loss.py
tbalke: > CI seems to be failing because the changes would slightly reduce the test coverage percentage.

Let's see whether that is still the case when we add the test for the Hessian.

scico/loss.py
bwohlberg: Understood, and that's a worthwhile goal, but we should make sure there aren't any undesirable consequences before we make this change.

scico/loss.py
bwohlberg: Perhaps you could remove them as part of this PR?

scico/loss.py
bwohlberg: Still unhappy, it seems. It would be best to address this before we merge.

smajee (changes requested):
What is the reasoning behind removing the diagonal op check in scico/linop/radon_svmbir.py ?
svmbir does not support non-diagonal weighing.

Otherwise looks good to me.

tbalke:
> What is the reasoning behind removing the diagonal op check in scico/linop/radon_svmbir.py ? svmbir does not support non-diagonal weighing.
> 
> Otherwise looks good to me.

The check is now in the `__init__` of `WeightedSquaredL2Loss` which now always requires a diagonal weight. The check is is performed at the `super().__init__(*args, **kwargs)` line.

scico/loss.py
Michael-T-McCann: I don't mind this change for the time being.

Longer term, consider: Why is `functional` a property at all if it is not used here? Broadly, I think this discussion is a symptom of the existence of `Loss`. One might hope to implement `__call__` at the level of `Loss` (and therefore using `self.functional`) to reduce repeated code. But we can't do that, because `Loss` is so general it doesn't know that it should be `A@x - y`.

Michael-T-McCann (approved):

scico/loss.py
bwohlberg: Agreed: if this change is made, we should consider removing the `functional` attribute. With respect to `Loss`, do you recall why it's so general? If there's good reason, perhaps we should have a specialization that really is `A@x - y`?

smajee (approved):

scico/loss.py
bwohlberg: I added a test for `loss.PoissonLoss` which will, I assume, resolve this.

@Michael-T-McCann : Would you not agree that the `Loss` tests should be in a separate `test_loss.py` file rather than included in `test_functional.py`?

scico/loss.py
```diff
@@ -273,17 +283,24 @@ def __init__(
...
         #: Constant term in Poisson log likehood; equal to ln(y!)
         self.const: float = gammaln(self.y + 1)  # ln(y!)
 
+        # The "true" Poisson loss is not differentiable at zero, but the
+        # ε that we add to allow evaluation at zero has the side-effect
+        # of making it differentiable there.
+        if isinstance(A, operator.Operator):
+            self.is_smooth = A.is_smooth
+        else:
+            self.is_smooth = None
```
bwohlberg, line 302: @Michael-T-McCann : Any thoughts on this addition? It's a bit ugly mathematically because we're labeling a non-smooth functional as smooth (because we smoothed it at zero), but this seems to be the right choice from a practical/computational perspective.

scico/loss.py
Michael-T-McCann: `Loss` is currently a subclass of `Functional` (was not always this way). Therefore I think it makes sense for the losses to get tested in `test_functional.py`, unless the file is much too long.

scico/loss.py
Michael-T-McCann: I think the reason it is general is because of the Poisson. I think the reason it exists at all is that it used to not be a subclass of `Functional` and therefore having a base `Loss` class made sense.

scico/loss.py
```diff
@@ -260,8 +271,7 @@ class PoissonLoss(Loss):
     Poisson negative log likelihood loss
 
     .. math::
-        \mathrm{scale} \cdot \sum_i [A(x)]_i - y_i \log\left( [A(x)]_i \right) + \log(y_i!)
-
+        \mathrm{scale} \sum_i [A(x)]_i - y_i \log\left( [A(x)]_i \right) + \log(y_i!)
```
Michael-T-McCann, line 274: Hate to say it: this looks odd to my eye. I think the cdot was intended to fix the oddness of `scale` being a word while the rest are math symbols. Elsewhere we have used `scale` for the name in code and \alpha in the math, see, e.g., https://scico--78.org.readthedocs.build/en/78/_autosummary/scico.functional.html#scico.functional.ScaledFunctional

scico/loss.py
Michael-T-McCann: My vote: document the use of epsilon, set `is_smooth=True`.

The only way around the epsilon I come up with is accounting for dark count rate in the forward model.

Michael-T-McCann (approved):

scico/loss.py
tbalke: I went with `\alpha`+`scaling parameter` in the docs, `scale` in the code now. (86e12ca)
@bwohlberg @Michael-T-McCann let me know what you think.

scico/loss.py
tbalke: Even with the epsilon it is still not smooth for `Ax < (-epsilon)`. I think a way to get around the mathematical imprecision is to say that the attribute we are looking for is whether it can be used in pgm or not.

If `A(x) = exp(x)` for example, then we can use the current `PoissonLoss` with no problem. But if `A(x)` is linear then we can even get into trouble when using the `epsilon` since `Ax < (-epsilon)` is very likely.

@bwohlberg is the `is_smooth` attribute used for anything other than the check in pgm?

scico/loss.py
bwohlberg: That looks much better.

scico/loss.py
bwohlberg: > Even with the epsilon it is still not smooth for `Ax < (-epsilon)`. 

Good point. Given that epsilon just shifts the problem, perhaps we should remove it? Alternatively, instead of adding epsilon, set values less than epsilon to epsilon?

> @bwohlberg is the `is_smooth` attribute used for anything other than the check in pgm?

I'm not sure. That may be the only use at the moment. Perhaps the whole `is_smooth` mechanism is worth a re-think. How is the issue of zeros handled in the Poisson loss example with PGM?

tbalke: > Given that epsilon just shifts the problem, perhaps we should remove it? Alternatively, instead of adding epsilon, set values less than epsilon to epsilon?

Yes, this may be a better option. But see below.

> Perhaps the whole is_smooth mechanism is worth a re-think.

Yes. I think if we rejected all non-smooth `L(y, A(x))` for all `x`, we would be unnecessarily impractical. A looser condition of `L(y, A(x))` smooth around `x_0` or in some feasible set could be more practical.

> How is the issue of zeros handled in the Poisson loss example with PGM?

(a) The initial condition is > 0 (almost surely).
(b) There is a non-negative indicator.
(c) Matrix A does not have negative entires
However, it seems like initializing to negatives or zero breaks the example code. I guess this kind of problem is not really that realistic. If we assume `y` to be Poisson, then in what world would `A(x)` be negative for any `x` in the feasible set.

I am leaning towards completely removing the epsilon and then then either `A` or the feasible set needs to be such that `A(x)>0`. (I think that currently would not break anything but I would not bet on it.)

@Michael-T-McCann ?

Michael-T-McCann: I had assumed the epsilon was necessary for some example or another. If it isn't, sure, remove it.

tbalke: Issue opened in #89

PR #79, opened 2021-11-09
Add function to compute operator norm

bwohlberg:
Add function to compute operator norm.

### DIFF
--- a/scico/linop/_linop.py
+++ b/scico/linop/_linop.py
@@ -35,17 +35,16 @@ def power_iteration(A: LinearOperator, maxiter: int = 100, key: Optional[PRNGKey
 
     Args:
         A: :class:`.LinearOperator` used for computation. Must be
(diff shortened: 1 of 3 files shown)


scico/linop/_linop.py
```diff
@@ -58,6 +57,40 @@ def power_iteration(A: LinearOperator, maxiter: int = 100, key: Optional[PRNGKey
     return mu, v
 
 
+def operator_norm(A: LinearOperator, maxiter: int = 100, key: Optional[PRNGKey] = None):
```
Michael-T-McCann, line 60: idea: it might be convenient to also have a norm property on `LinearOperator`s that can be set during construction if it is easy to compute. Circulant convolutions come to mind. Then optimizers that want the norm can see if `A.norm is not None` before running the potentially expensive power iteration.

Michael-T-McCann (approved):
good to merge, my comment not meant as a blocker

PR #87, opened 2021-11-11
Consistent notation for Hermitian conjugate

bwohlberg:
In relevant equations, replace `A^*` with `A^H` to denote the Hermitian conjugate.


Michael-T-McCann (approved):
Fine with me. I would usually write A^*, but I if we have ^H most places let's stick with it.

bwohlberg:
> Fine with me. I would usually write A^*, but I if we have ^H most places let's stick with it.

The `A^H` form may have slightly more usage, but to me the more compelling argument is the ambiguity of `A^*` in that it could also be taken to mean just the complex conjugate.

lukepfister:
A little late on this one, but here was the original thinking:  

- `^H` should be reserved for finite-dimensional things (matrices!)
- `^*` is more general and is used for linear operators between generic spaces

i doubt i was actually consistent with that, though.

Michael-T-McCann:
Even later to this: is there any reasonable argument that A* should denote elementwise complex conjugation? I have never seen it used this way.

bwohlberg:
It's frequently used to denote the complex conjugate of a scalar variable, and notation for scalar operations is often applied to vectors and matrices to indicate elementwise operations, so that wouldn't seem like an unreasonable possible interpretation.

Issue #57, opened 2021-10-21
Additional tests for svmbir interface


bwohlberg:
Additional tests needed to confirm that the svmbir prox solver gives the same results as a CG solve of the prox problem involving the svmbir forward and adjoint operators.

bwohlberg:
ToDo:
- [ ]  Add an initializer parameter (default None) for `Functional.prox`
- [ ] Add passing of appropriate initializer to prox in `ADMM` (and `PGM`?)
- [ ] Add tests based on operators and CG for prox of `SVMBIRWeightedSquaredL2Loss`
- [ ] Initializers for `radon_svmbir.ParallelBeamProjector` and `SVMBIRWeightedSquaredL2Loss` should set mask radius to sufficient size to avoid masking and provide parameters that allow setting of svmbir prox solver stopping tolerance and max steps
- [ ] Choose appropriate defaults for svmbir prox solver stopping tolerance and max steps

Michael-T-McCann:
In branch `mike/approx_prox` proxes all take `**kwargs` and ADMM passes an initial value for the prox as `v0`.

PR #88, opened 2021-11-12
Changes to admm and svmbir wrapper to address issue #57

smajee:
Addresses issue #57 

Overview of changes (Mike and Soumendu):
- Pass initializer to prox in admm
- Use initial condition in svmbir prox
- Make number of iterations in svmbir prox user-defined (with large number of iterations as default)
- Make roi radius large by default in svmbir so that there is no masking and the solution matches with CG. (masking can be enabled by the user if required)
- Use nanfix branch of svmbir in requirements (we can change it back once the fix is pushed to PyPi)


bwohlberg:
Does this address all of the remaining check boxes (other than the PGM one) in #57?

smajee:
> Does this address all of the remaining check boxes (other than the PGM one) in #57?

This doesn't contain the tests for CG and SVMBIRWeightedSquaredL2Loss.

svmbir_test.py
```diff
@@ -0,0 +1,99 @@
+import matplotlib.pyplot as plt
```
tbalke, line 1: This file should be deleted before merge. It is just for testing purposes.

scico/test/linop/test_radon_svmbir.py
```diff
@@ -40,6 +42,24 @@ def make_A(im, num_angles, num_channels):
     return A
 
 
+def cg_prox(f, v, λ):
```
tbalke, line 45: @bwohlberg @Michael-T-McCann This prox implementation may be useful. I could add this to WeightedSquaredL2Loss and SquaredL2Loss so we have a default even when A is not diagonal.

scico/functional/_functional.py
```diff
@@ -252,5 +260,7 @@ class ZeroFunctional(Functional):
     def __call__(self, x: Union[JaxArray, BlockArray]) -> float:
         return 0.0
 
-    def prox(self, x: Union[JaxArray, BlockArray], lam: float = 1) -> Union[JaxArray, BlockArray]:
+    def prox(
+        self, x: Union[JaxArray, BlockArray], lam: float = 1, **kwargs
```
tbalke, line 264: 1.0 here

scico/linop/radon_svmbir.py
```diff
@@ -39,6 +41,7 @@ def __init__(
         input_shape: Shape,
         angles: np.ndarray,
         num_channels: int,
+        is_masked=False,
```
tbalke, line 44: Add docstring

scico/linop/radon_svmbir.py
```diff
@@ -39,6 +41,7 @@ def __init__(
...
     ):
         """
         Args:
```
tbalke, line 47: More explanation on the arguments would be nice

tbalke:
> > Does this address all of the remaining check boxes (other than the PGM one) in #57?
> 
> This doesn't contain the tests for CG and SVMBIRWeightedSquaredL2Loss.

Now it does.

scico/functional/_functional.py
smajee: Done

scico/linop/radon_svmbir.py
smajee: Done

scico/linop/radon_svmbir.py
smajee: Done

requirements.txt
```diff
@@ -6,4 +6,5 @@ jax==0.2.19
 jaxlib==0.1.70
 flax
 bm3d
-svmbir
+# svmbir
+git+https://github.com/cabouman/svmbir.git@nanfix
```
bwohlberg, line 10: When this is merged, please add an issue as a reminder to revert this to just `svmbir` as soon as the fix is released.

scico/linop/radon_svmbir.py
```diff
@@ -24,12 +32,6 @@
         "for instructions on how to install the SVMBIR."
```
bwohlberg, line 32: For grammatical and consistency reasons, change to "... install svmbir", or even better, simply "... how to install it."

scico/linop/radon_svmbir.py
```diff
@@ -39,13 +41,18 @@ def __init__(
...
         """
         Args:
             input_shape: Shape of the input array.
             angles: Array of projection angles in radians, should be
               increasing.
             num_channels: Number of pixels in the sinogram
+            is_masked: If True, valid region of the
+                image is set to a circular region prescribed in the
+                image array. Otherwise, the whole image array is valid
+                for projection and reconstruction.
```
bwohlberg, line 55: "prescribed in the image array" is a little opaque.

scico/linop/radon_svmbir.py
```diff
@@ -63,6 +70,11 @@ def __init__(
                 f"Only 2D and 3D inputs are supported, but input_shape was {input_shape}"
             )
 
+        if is_masked:
+            self.roi_radius = None
```
bwohlberg, line 74: Tests need to cover both possible values of `is_masked`.

scico/linop/radon_svmbir.py
```diff
@@ -111,7 +139,7 @@ def _bproj_hcb(self, y):
 
 
 class SVMBIRWeightedSquaredL2Loss(WeightedSquaredL2Loss):
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, max_iterations=1000, stop_threshold=0.001, **kwargs):
         super().__init__(*args, **kwargs)
```
bwohlberg, line 143: It looks as if `max_iterations` and `stop_threshold` are just discarded: they're just passed to the `__init__` of `WeightedSquaredL2Loss`, which ignores them.

scico/linop/radon_svmbir.py
```diff
@@ -122,23 +150,39 @@ def __init__(self, *args, **kwargs):
...
+        self.max_iterations = max_iterations
+        self.stop_threshold = stop_threshold
+
+    def prox(self, v: JaxArray, lam: float, **kwargs) -> JaxArray:
         v = v.reshape(self.A.svmbir_input_shape)
         y = self.y.reshape(self.A.svmbir_output_shape)
         weights = self.W.diagonal.reshape(self.A.svmbir_output_shape)
         sigma_p = snp.sqrt(lam)
+        if "v0" in kwargs and kwargs["v0"] is not None:
+            v0 = np.reshape(np.array(kwargs["v0"]), self.A.svmbir_input_shape)
```
bwohlberg, line 162: Need to include case with `v0` provided in tests.

scico/test/linop/test_radon_svmbir.py
bwohlberg: Agreed. Please do so. It should also take `**kwargs` so that `v0` can be specified, and there should also be some way of specifying max iterations and stopping tolerance. Perhaps these should go in the `__init__`? @Michael-T-McCann, thoughts?

scico/test/linop/test_radon_svmbir.py
```diff
@@ -40,6 +42,24 @@ def make_A(im, num_angles, num_channels):
...
+    # prox:
+    #   arg min  1/2 || x - v ||^2 + λ α || A x - y ||^2_W
+    #      x
+    #
+    # solution at:
+    #   (I + λ 2α A^T W A) x = v + λ 2α A^T W y
```
bwohlberg, line 51: Missing square root: should be A^T W^(1/2) y

scico/test/linop/test_radon_svmbir.py
```diff
@@ -40,6 +42,24 @@ def make_A(im, num_angles, num_channels):
...
+    W = f.W
+    A = f.A
+    α = f.scale
+    y = f.y
+    hessian = f.hessian  # = (2α A^T W A)
+    lhs = Identity(v.shape) + λ * hessian
+    rhs = v + 2 * λ * α * A.adj(W(y))
```
bwohlberg, line 58: Square root missing here too.

scico/test/linop/test_radon_svmbir.py
```diff
@@ -92,3 +112,29 @@ def test_prox_weights(Nx, Ny, num_angles, num_channels, is_3d):
...
+def test_prox_cg(Nx, Ny, num_angles, num_channels, is_3d, is_weighted):
+    im = make_im(Nx, Ny, is_3d=is_3d) / Nx * 10
+    A = make_A(im, num_angles, num_channels)
+    y = A @ im
+
+    if is_weighted:
+        W = snp.exp(-y) * 20
+    else:
+        W = snp.ones_like(y)
+
```
bwohlberg, line 129: Also support `W=None`?

svmbir_test.py
```diff
@@ -0,0 +1,99 @@
...
+    #
+    # solution at:
+    #   (I + λ 2α A^T W A) x = v + λ 2α A^T W y
+    W = f.W
+    A = f.A
+    α = f.scale
+    y = f.y
+    hessian = f.hessian  # = (2α A^T W A)
+    lhs = Identity(v.shape) + λ * hessian
+    rhs = v + 2 * λ * α * A.adj(W(y))
```
bwohlberg, line 43: Note missing square root.

svmbir_test.py
bwohlberg: Perhaps best to mark the PR as a draft until this file has been deleted.

bwohlberg:
Please fill in ToDo checkboxes in #57 so we can see how close this PR is to completely resolving it.

scico/test/linop/test_radon_svmbir.py
tbalke: You sure? With the new definition of W I think it is only W, not W^(1/2)

`||y||_W^2 = y^T W y`

bwohlberg: Yes, you're right - the perils of late night mental algebra! This also needs to be corrected in  docstrings and code merged from `brendt/admm-bug`.

bwohlberg: This should now have been fixed by another merge with a corrected `brendt/admm-bug`.

scico/linop/radon_svmbir.py
smajee: How about:
 If True, valid region of the
image is set to the largest circle that is prescribed
within the image array.
Otherwise, the whole image array is valid
for projection and reconstruction.

scico/linop/radon_svmbir.py
smajee: Done

scico/linop/radon_svmbir.py
smajee: In line 153-154 I use them:
`self.max_iterations = max_iterations`
`self.stop_threshold = stop_threshold`.
These are then used in the recon call.
The `__init__` of `WeightedSquaredL2Loss` just gets the rest of the arguments.

scico/linop/radon_svmbir.py
smajee: I added tests that cover both values of `is_masked`. (except `test_prox_cg' since we don't expect it to pass if there is masking)

scico/linop/radon_svmbir.py
smajee: But we could alternatively pass these to the prox directly.
What do you think?

scico/linop/radon_svmbir.py
bwohlberg: I suggest: If True, the valid region of the image is determined by a mask defined as the circle inscribed within the image boundary.  Otherwise ....

Is it clear from context that "image" refers to the reconstruction and not the sinogram?

scico/linop/radon_svmbir.py
bwohlberg: There's no way of making a working `test_prox_cg` with masking enabled?

scico/linop/radon_svmbir.py
bwohlberg: Oh, for some reason I didn't see that. No, probably best to leave them where they are.

scico/linop/radon_svmbir.py
smajee: I added a test for `v0`. (Checks if true prox is same as approximate prox with a good initialization).

scico/linop/radon_svmbir.py
smajee: Made the changes. I added more context so that it is clear what 'image' refers to.

scico/linop/radon_svmbir.py
smajee: One way (although a little hacky) could be to find the mask either by back-projecting a sinogram of 1's or by thresholding a reconstruction and then use the mask for comparing the norms of the two proxes. 

Let me know whether this sounds reasonable.

scico/test/linop/test_radon_svmbir.py
smajee: Would it make more sense to use the svmbir.calc_weights() here since that is what the examples use?

smajee: We could then pass the various strings ("unweighted", "transmission" etc) as the test argument

requirements.txt
```diff
@@ -7,4 +7,4 @@ jaxlib==0.1.70
 flax
 bm3d
 # svmbir
-git+https://github.com/cabouman/svmbir.git@nanfix
\ No newline at end of file
+git+https://github.com/cabouman/svmbir.git@master
```
bwohlberg, line 10: I expect the release will be on PyPI quite soon. When it does, this should become `svmbir>=0.2.6`.

scico/test/linop/test_radon_svmbir.py
bwohlberg: Good idea.

scico/linop/radon_svmbir.py
bwohlberg: It's perhaps a bit hacky for the main package code, but that would be fine for a test.

smajee:
> Please fill in ToDo checkboxes in #57 so we can see how close this PR is to completely resolving it.

Done

requirements.txt
bwohlberg: It's on PyPI now, so this change no longer has to be delayed.

smajee: Done

scico/linop/radon_svmbir.py
smajee: I expanded the test to include the masked case. For the masked case, the comparison between cg and inbuilt-prox is done in the masked region. The mask is computed by back-projecting 1's.

However, I noticed that for the small input case, the mask is the full image. Should we have a larger test case?

Another thing I noticed was that svmbir.prox returns ndarray whereas cg returns jax-device array. Is this by design, or shall we make them both the same data type?

scico/test/linop/test_radon_svmbir.py
smajee: Updated the test

scico/linop/radon_svmbir.py
```diff
@@ -11,41 +11,51 @@
...
-from scico.loss import WeightedSquaredL2Loss
-from scico.typing import JaxArray, Shape
-
-from ._linop import LinearOperator
-
 
 class ParallelBeamProjector(LinearOperator):
-    r"""Parallel beam projector based on svmbir."""
+    r"""Parallel beam projector based on svmbir.
+    Performs tomographic projections of an image at specified angles.
```
bwohlberg, line 38: Nitpicks:
1. There should be a blank line after the initial summary line.
2. Phrasing should typically be imperative, i.e. "Perform tomographic ...".

scico/linop/radon_svmbir.py
```diff
@@ -11,41 +11,51 @@
...
+    """
 
     def __init__(
         self,
         input_shape: Shape,
         angles: np.ndarray,
         num_channels: int,
+        is_masked: Optional[bool] = False,
```
bwohlberg, line 46: Also worth exposing the non-negativity option of svmbir here?

scico/linop/radon_svmbir.py
bwohlberg: > However, I noticed that for the small input case, the mask is the full image. Should we have a larger test case?

I think it's fine as-is.

> Another thing I noticed was that svmbir.prox returns ndarray whereas cg returns jax-device array. Is this by design, or shall we make them both the same data type?

I would expect consistency in data type, but perhaps there's a good reason for this choice. @Michael-T-McCann and @tbalke, can you comment on this?

svmbir_test.py
bwohlberg: Is this still needed? It looks as if the PR is almost ready to go, so perhaps it should be deleted?

scico/linop/radon_svmbir.py
smajee: I think `SVMBIRWeightedSquaredL2Loss` would be more appropriate for adding the non-negativity since non-negativity does not have any bearing on projection and back-projection. Let me know what you think.

bwohlberg: Yes, you're right - I didn't look closely and had thought my comment was with respect to that class.

scico/linop/radon_svmbir.py
smajee: Made the change

scico/linop/radon_svmbir.py
smajee: Exposed positivity to the user.

scico/test/linop/test_radon_svmbir.py
tbalke: We have the following candidates for arguments that may want to get passed to the cg-prox:
```python
    x0: JaxArray,
    tol: float = 1e-5,
    atol: float = 0.0,
    maxiter: int = 1000,
    M: Optional[Callable] = None,
```
Perhaps we want to keep it consistent with the SVMBIR-prox and only use 
```python
max_iterations=1000,
stop_threshold=0.001
v0=None
```
with these variable names?

tbalke: @bwohlberg  I started a draft of these changes in 3fdc0cb.

- This uses the cg from the `WeightedSquaredL2Loss` to test against the SVMBIR prox. What do you think?
- It does not use any max_iterations or tolerance arguments yet until we agree how to best do this consistently
- It does not have the prox for the unweighted loss yet. Will do once we agree on the weighted version.

svmbir_test.py
tbalke: Currently still using it. (Can remove it from the repo though if you prefer)

bwohlberg: No, we should only delete it once it's no longer needed.

scico/linop/radon_svmbir.py
tbalke: @smajee We recently implemented the `ensure_on_device` routine when we realized that we need to restrict the _input_ types somewhat. (Making it work with ndarrays became impractical). Our conclusion was that we build in some safeguards but in the end the user has to have some foresight since unnecessary conversion may lead to performance slowdowns. 

I think the current stage on what type our routines _return_ is: they may return ndarrays, device arrays, or block arrays but we don't seem to be very precise and consistent about it. I think most routines will not return nd-arrays though (maybe svmbir stuff is the only exception) because once on device, always on device.

So that makes me lean towards casting the svmbir output to device array. Especially, when considering that the prox may return an ndarray somewhere deep inside the admm iterations, it seem to break our assumption that the optimizers are internally only working with device arrays.

bwohlberg: At some point we should discuss whether it's possible to be more consistent on return types.

The `svmbir` and `bm3d` interfaces are special cases in that they can't be run a GPU, but I agree with your concern about prox returning an ndarray in a context in which device array is expected. My immediate response would be that making the output type the same as the input is a good approach, but this deserves a bit more consideration. (And we should make sure that the same approach is followed for the `bm3d` interface.)

scico/linop/radon_svmbir.py
```diff
@@ -11,41 +11,52 @@
...
     ):
         """
         Args:
             input_shape: Shape of the input array.
             angles: Array of projection angles in radians, should be
-              increasing.
+                increasing.
             num_channels: Number of pixels in the sinogram
+            is_masked:  If True, the valid region of the image is
```
Michael-T-McCann, line 55: Not obvious to me what "valid region" means. Is something set to zero? When?

scico/linop/radon_svmbir.py
Michael-T-McCann: I don't think there is a special reason to return an `ndarray`. Agreed that matching the input type makes sense.

scico/linop/radon_svmbir.py
smajee: How about:

is_masked:  If True, the valid region of the image is
                determined by a mask defined as the circle inscribed
                within the image boundary.
                Otherwise, the whole image array is valid
                for projection and reconstruction.
                Pixels outside the valid region is not projected or updated during the reconstruction.

smajee: Updated doc-string to make it clearer

bwohlberg:
Recent changes in `scico.loss` (presumably the addition of the CG-based proxes) seems to have introduced quite a bit of untested code (see codecov report). It would be best to add the tests as part of this PR.

tbalke:
> Recent changes in `scico.loss` (presumably the addition of the CG-based proxes) seems to have introduced quite a bit of untested code (see codecov report). It would be best to add the tests as part of this PR.

Passing now !

tbalke:
Added some tests for the new proxes in 

> Recent changes in `scico.loss` (presumably the addition of the CG-based proxes) seems to have introduced quite a bit of untested code (see codecov report). It would be best to add the tests as part of this PR.

Added some tests for the proxes in 53713dd. Not enough.

tbalke (approved):

scico/linop/radon_svmbir.py
```diff
@@ -111,35 +145,58 @@ def _bproj_hcb(self, y):
...
             num_cols=self.A.svmbir_input_shape[2],
+            roi_radius=self.A.roi_radius,
             sigma_p=float(sigma_p),
             sigma_y=1.0,
-            positivity=False,
+            positivity=self.positivity,
             verbose=0,
+            init_image=v0,
+            max_iterations=self.prox_kwargs.get("maxiter"),
+            stop_threshold=self.prox_kwargs.get("ctol"),
```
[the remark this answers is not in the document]
bwohlberg: Agreed. Any suggestions?

scico/linop/radon_svmbir.py
```diff
@@ -111,7 +143,9 @@ def _bproj_hcb(self, y):
 
 
 class SVMBIRWeightedSquaredL2Loss(WeightedSquaredL2Loss):
```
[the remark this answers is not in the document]
bwohlberg: Yes. That's a pretty big oversight. It needs to be fixed before we merge.

More generally, we should try to expand the workflow lint tests to catch cases like this.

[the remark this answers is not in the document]
bwohlberg: @Michael-T-McCann and @smajee : Could you look at the missing docstring issue?

examples/scripts/ct_svmbir_ppp_bm3d_admm_prox.py
```diff
@@ -91,7 +91,9 @@
 ρ = 10  # ADMM penalty parameter
 σ = density * 0.26  # denoiser sigma
 
-f = SVMBIRWeightedSquaredL2Loss(y=y, A=A, W=Diagonal(weights), scale=0.5)
+f = SVMBIRWeightedSquaredL2Loss(
+    y=y, A=A, W=Diagonal(weights), scale=0.5, prox_kwargs={"maxiter": 5}
```
bwohlberg, line 95: Is it really desirable that we just drop the default stopping tolerance here?

scico/linop/radon_svmbir.py
```diff
@@ -11,41 +11,52 @@
...
 class ParallelBeamProjector(LinearOperator):
-    r"""Parallel beam projector based on svmbir."""
+    r"""Parallel beam Radon transform based on svmbir.
+
+    Perform tomographic projection of an image at specified angles,
+    using the `svmbir <https://github.com/cabouman/svmbir>`_ package. The
+    ``is_masked`` option defines a valid region for projection and updates.
+    Pixels outside the valid region are not projected or updated during
+    a reconstruction.
+    """
```
bwohlberg, line 41: It's not clear what `updates` and `reconstruction` here refer too, since this class is just a Radon transform. Presumably it refers to the when the `ParallelBeamProjector` is used with `SVMBIRWeightedSquaredL2Loss`? If so, this should be clarified here.

examples/scripts/ct_svmbir_ppp_bm3d_admm_prox.py
tbalke: I think I would choose the tolerance to be 0 (or very) low so that always 5 iterations are being performed (unless there is really almost no change)

scico/linop/radon_svmbir.py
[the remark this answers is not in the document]
tbalke: Working on right now

scico/linop/radon_svmbir.py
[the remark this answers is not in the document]
tbalke: Clarified this in 079ac16

scico/linop/radon_svmbir.py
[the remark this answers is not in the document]
tbalke: Addressed in 079ac16

examples/scripts/ct_svmbir_ppp_bm3d_admm_prox.py
tbalke: -> a640fa7

svmbir_test.py
tbalke: Done

scico/linop/radon_svmbir.py
smajee: Clarified in commit 219e5ed

bwohlberg (approved):

PR #99, opened 2021-11-16
Skip test based on content of scico/data/examples

### LINKED ISSUE #98: Tests for scico.data fail when run on installed package
The tests for `scico.data` fail if run on an installed version of `scico`
```
pytest --pyargs scico
```
because the test skip conditional assumes the test is being run from the distribution root directory, and fails when no `data` directory is found.

bwohlberg:
Skip test based on content of `scico/data/examples` rather than content of of `data` directory.

Still to be confirmed that this modified skip test still achieves the original goal of skipping the tests for `scico.data` when run within a cloned repo without the `scico-data` submodule present.

Resolves #98.

### DIFF
--- a/scico/test/test_data.py
+++ b/scico/test/test_data.py
@@ -12,9 +12,8 @@ skipif_reason = (
     "\nAnd after cloning run:\n\tgit submodule init && git submodule update.\n"
 )
 
-pytestmark = pytest.mark.skipif(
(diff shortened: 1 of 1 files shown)


FernandoDavis (approved):
Changed skipif condition to match general case. Works as expected in a new repo.

Issue #103, opened 2021-11-17
The curious case of the ASTRA adjoint on GPU


Michael-T-McCann:
The ASTRA Toolbox GPU back projector is known to not be a highly accurate adjoint to the forward operator. In the old days, we handled this with a generous tolerance on our adjoint test. After some improvements to that adjoint test, the ASTRA back projector fails _very_ hard (relative error 0.78).

The question: is this a problem with the test, our interface to ASTRA, or ASTRA itself? If it's the latter, let's just skip these tests on GPU.

First step: get a GPU install of SCICO and run `pytest scico/test/linop/test_radon_astra.py` to see if the failures occur for you.

lukepfister:
IIRC it’s a problem with the astra algorithm itself.

smajee:
So if it is due to the inaccuracy of the astra projector, should we increase the tolerance of the test or just disable them for gpu?

Michael-T-McCann:
To clarify: we know there is _at least_ a problem with the ASTRA GPU backprojection algorithm itself. The question here why is the new adjoint test failing with a relative error approaching one. I didn't think the ASTRA backprojection was _that_ inaccurate, but I don't have a good intuition for our new adjoint test.

If you are convinced there is no bug in the interface or test, I learn toward increasing the tolerance rather than disabling the test, because it might allow us to catch regressions in the interface in the future.

lukepfister:
has the astra version changed? it looks like they finally shipped v2 (was on 1.9.9 beta for a looong time)

has the jax/jaxlib version changed?  might be worth seeing if something has been modified in the host_callback interface;  it is marked as "experimental" and subject to breaking changes

smajee:
Is the adjoint test new? Did it ever pass on GPU for astra?

lukepfister:
No, it isn't new.  It passed with very, very loose tolerances; like a relative error of 20% or something.

Sounds like that's no longer the case (and I have no gpu machine to test)

smajee:
I did pytests on old commits to see where it broke on GPU.
It seems the tests pass on commit d86b51f and fail after commit 656f59d (Improve LinearOperator adjoint tests (#72)).

That commit seems to be the one where a new adjoint test (`adjoint_test`) was introduced. The old `adjoint_AAt_test` seems to pass on GPU.

I don't understand the tests well enough to comment on where why one passes and the other fails. Any ideas?

Michael-T-McCann:
[...]

PR #72, opened 2021-10-30
Improve LinearOperator adjoint tests

bwohlberg:
Improve LinearOperator adjoint tests. Resolves #66.

PR #107, opened 2021-11-18
Fix ASTRA test in GPU issue #103

smajee:
I did pytests on old commits to see where it broke on GPU.
It seems the tests pass on commit d86b51f and fail after commit 656f59d (Improve LinearOperator adjoint tests (#72)).

That commit (656f59d) seems to be the one where a new adjoint test (adjoint_test) was introduced. The old adjoint_AAt_test seems to pass on GPU.

Based on the discussion in  issue #103 I increased the tolerance for the `adjoint_test`.


scico/test/linop/test_radon_astra.py
```diff
@@ -95,4 +95,4 @@ def test_adjoint_grad(testobj):
 
 def test_adjoint(testobj):
     A = testobj.A
-    adjoint_test(A, rtol=get_tol())
+    adjoint_test(A, rtol=15 * get_tol())
```
Michael-T-McCann, line 98: I'm now skeptical of the use of relative error in this test.

I believe the test boils down to relative error = |<y, Ax> - <Aty, x>| / max(<y, Ax>, <Aty, x>). But what if, e.g., A represents a 90 degree rotation? Then all these quantities are (nearly) zero and the resulting relative error is undefined/unstable.

scico/linop/_linop.py
```diff
@@ -132,13 +132,13 @@ def valid_adjoint(
...
-    x1y0 = snp.dot(x1.ravel().conj(), y0.ravel())
-    y1x0 = snp.dot(y1.ravel().conj(), x0.ravel())
-    err = snp.linalg.norm(x1y0 - y1x0) / max(snp.linalg.norm(x1y0), snp.linalg.norm(y1x0))
+    x, key = randn(shape=A.input_shape, key=key, dtype=A.input_dtype)
+    y, key = randn(shape=AT.input_shape, key=key, dtype=AT.input_dtype)
+    u = A(x)
+    v = AT(y)
+    yTu = snp.dot(y.ravel().conj(), u.ravel())
+    vTx = snp.dot(v.ravel().conj(), x.ravel())
+    err = snp.linalg.norm(yTu - vTx) / max(snp.linalg.norm(yTu), snp.linalg.norm(vTx))
```
tbalke, line 141: This is actually not what the docstring claims. The docstring uses absolute error, however the code uses relative error.

scico/test/linop/test_radon_astra.py
smajee: That is a good point. In the astra test context, it computes the dot product between a sinogram and the image. Given they are in two domains, their dot-product can be small. This doesn't make seem reasonable to me in a tomography context.

In the CT context we can divide by the norm of the image instead for the relative error. But not sure if that makes sense in the other cases the adjoint test is applied to.

tbalke: > I believe the test boils down to relative error = |<y, Ax> - <Aty, x>| / max(<y, Ax>, <Aty, x>). But what if, e.g., A represents a 90 degree rotation? Then all these quantities are (nearly) zero and the resulting relative error is undefined/unstable.

Sure about that? `<y, Ax>` is only small if `y` and `Ax` happen to be close to perpendicular (or if `A` is small but then also the numerator will be small).

> In the astra test context, it computes the dot product between a sinogram and the image.

Really? Seems like the dot product between _**a random**_ sinogram (`y`) and _**a random**_ projected image (`Ax`).

scico/linop/_linop.py
bwohlberg: Good point. It would make sense to keep using relative error in the implementation and to update the docs to match.

tbalke: Will do.

bwohlberg:
> I'm now skeptical of the use of relative error in this test.
> ...
>  Then all these quantities are (nearly) zero and the resulting relative error is undefined/unstable.

The undefined case (i.e. quantities are actually zero) can be handled as in [sporco.linalg.rrs](https://sporco.readthedocs.io/en/latest/_modules/sporco/linalg.html#rrs). I haven't given careful thought to the instability when they're just very small, but my inclination is that it's a reasonable proce to pay for the benefits of relative vs. absolute error.

scico/test/linop/test_radon_astra.py
smajee: By doing this way, all the tests pass.

bwohlberg: What is the actual numerical value of ``15 * get_tol()``? Performing a test like this with a relative error tolerance close to unity seems somewhat pointless.

scico/linop/_linop.py
tbalke: d67bc23

scico/linop/_linop.py
```diff
@@ -107,8 +107,10 @@ def valid_adjoint(
 
     by computing :math:`\mathbf{u} = A \mathbf{x}` and
     :math:`\mathbf{v} = A^T \mathbf{y}` for random :math:`\mathbf{x}`
-    and :math:`\mathbf{y}` and confirming that :math:`\| \mathbf{y}^T
-    \mathbf{u} - \mathbf{v}^T \mathbf{x} \|_2 < \epsilon` since
+    and :math:`\mathbf{y}` and confirming that :math:`\frac{\| \mathbf{y}^T
+    \mathbf{u} - \mathbf{v}^T \mathbf{x} \|_2}{\max \left\{
+    \| \mathbf{y}^T \mathbf{u} \|_2, \| \mathbf{v}^T \mathbf{x} \|_2
+    \right\}} < \epsilon` since
```
bwohlberg, line 113: The docstring looks a bit strange like this. I suggest displaying the equation containing the ``\frac{}{}``.

tbalke: Not really sure what you mean. Can you elaborate, please?

scico/test/linop/test_radon_astra.py
Michael-T-McCann: `15 * get_tol()` is 1.05. That makes it appear that the ASTRA backprojector is very inaccurate, but it's not the case as far as I know. Hence my skepticism of relative error.

Michael-T-McCann: @tbalke good point, my 90 rotation example doesn't make sense because it is _not_ the case that y = Ax here.

smajee: > > I believe the test boils down to relative error = |<y, Ax> - <Aty, x>| / max(<y, Ax>, <Aty, x>). But what if, e.g., A represents a 90 degree rotation? Then all these quantities are (nearly) zero and the resulting relative error is undefined/unstable.
> 
> Sure about that? `<y, Ax>` is only small if `y` and `Ax` happen to be close to perpendicular (or if `A` is small but then also the numerator will be small).
> 
> > In the astra test context, it computes the dot product between a sinogram and the image.
> 
> Really? Seems like the dot product between _**a random**_ sinogram (`y`) and _**a random**_ projected image (`Ax`).

Sorry, I was mistaken about that.
But I still think normalizing by the norm of the images instead of the dot product can be a better option for the relative tolerance. That way the denominator wouldn't get close to zero.

smajee: Based on @Michael-T-McCann 's suggestion, I plotted  E[|<x,y>| / <x,x>]) vs number of dimensions where x,y are random normal vectors.

Seems to drop to zero pretty fast. So normalizing by the norm of the image rather than the dot product in the test can be a better normalization.

Screen Shot 2021-11-19 at 2 57 29 PM

smajee: Made changes to the normalization in commit e894d35. Let me know your thoughts on this. (haven't updated the docstring though).

scico/linop/_linop.py
bwohlberg: I meant that it would be better to write using the ``.. math::`` form than the inline ``:math:`` form.

scico/test/linop/test_radon_astra.py
bwohlberg: > ``15 * get_tol()`` is 1.05. That makes it appear that the ASTRA backprojector is very inaccurate, but it's not the case as far as I know. Hence my skepticism of relative error.

I agree that it doesn't look good for the ASTRA backprojector, but I'm not quite ready to just chalk this up as a failing of relative error.

bwohlberg: > Seems to drop to zero pretty fast. So normalizing by the norm of the image rather than the dot product in the test can be a better normalization.

> Made changes to the normalization in commit e894d35. Let me know your thoughts on this. (haven't updated the docstring though).

I'm not a fan of this approach because it doesn't normalize for changes in the norm of `A`, so a simple scaling of `A` would change our measure of adjoint accuracy. (But if one does take this approach, I think it would be better to just normalize the random `x` and `y` vectors after they're created.)

scico/linop/_linop.py
tbalke: I made that and a couple changes in 76337c6.

- clearly different font distinction between matrix (A, A^T) and operators (A, AT)
- removed remedial equation after 'since ...'
- display-equation for the error inequality

scico/test/linop/test_radon_astra.py
smajee: I think it does normalize for changes in the norm of A. 

In commit e894d35 I normalized by max(||y||^2, ||Ax||^2). So scaling A would change both numerator and denominator for the relative error. But doing the max() may not be ideal. A better way might be normalizing by (||y|| ||Ax||)

bwohlberg: I didn't notice that in the commit; I thought you were just normalizing by the max of the norms of x and y. The max is really problematic because the scaling of the denominator with A depends on whether the norm of y is smaller than the norm of Ax. Normalizing by their product resolves that problem, but I'm not convinced that it's better than the original form, which seems to me to be more easily interpretable.

smajee: The original form essentially normalizes by |y^T Ax|.
The problem with this is that for large random vectors y and Ax, the dot product can be fairly small since x,y are zero mean Gaussian vectors. This can make the relative error large. This issue wouldn't occur if x and y are constrained to be non-negative.

smajee: Added new commit e2130b0

scico/linop/_linop.py
```diff
@@ -97,33 +97,35 @@ def valid_adjoint(
...
-    \mathbf{u} - \mathbf{v}^T \mathbf{x} \|_2 < \epsilon` since
+    by computing :math:`\mathbf{u} = \mathsf{A}(\mathbf{x})` and
+    :math:`\mathbf{v} = \mathsf{AT}(\mathbf{y})` for random :math:`\mathbf{x}`
+    and :math:`\mathbf{y}` and confirming that
 
     .. math::
-      \mathbf{y}^T \mathbf{u} = \mathbf{y}^T (A \mathbf{x}) =
-      (A^T \mathbf{y})^T \mathbf{x} = \mathbf{v}^T \mathbf{x}
+      \frac{\| \mathbf{y}^T \mathbf{u} - \mathbf{v}^T \mathbf{x} \|_2}
+      {\max \left\{ \| \mathbf{y}^T \mathbf{u} \|_2,
```
Michael-T-McCann, line 114: I suggest using absolute value rather than norm as these are scalar quantities.

scico/linop/_linop.py
```diff
@@ -97,33 +97,35 @@ def valid_adjoint(
...
+    test as above, but with all the :math:`(\cdot)^T` replaced with
+    :math:`(\cdot)^H`.
 
     Args:
         A: Primary :class:`.LinearOperator`.
         AT: Adjoint :class:`.LinearOperator`.
-        eps: Error threshold for validation of `AT` as adjoint of `A`. If
-           None, the relative error is returned instead of a boolean value.
+        eps: Error threshold for validation of :math:`\mathsf{AT}` as
+           adjoint of :math:`\mathsf{AT}`. If None, the relative error
```
Michael-T-McCann, line 127: Mention that the default is 1e-7.

bwohlberg: Some of the older scico code does this, but more recent practice seems to expect the user to inspect the function signature in the docs to determine defaults. Yet another minor thing to discuss.

Michael-T-McCann: I didn't see it in the signature, no change needed.

scico/linop/_linop.py
smajee: I think I fixed it. Can someone double check? I seem to be having issues with building the docs.

scico/test/linop/test_radon_astra.py
smajee: Added the changes we discussed in the meeting.

scico/linop/_linop.py
Michael-T-McCann: The CI builds the docs automatically, look below this comment, "show all checks" -> docs/readthedocs.org:scico, click "details"

Michael-T-McCann: They still look like norms, need to change `\|` to `|` or `\lvert` `\rvert`.

smajee: > The CI builds the docs automatically, look below this comment, "show all checks" -> docs/readthedocs.org:scico, click "details"

Thanks. Good to know. Made the change using `|`

bwohlberg (approved):

PR #108, opened 2021-11-20
Clean up notebooks build script

bwohlberg:
Improve Python notebook build script, including removal of the need to exec a bash script. Remove makefile and bash script.


examples/README.rst
```diff
@@ -40,17 +40,11 @@ A number of files in this directory assist in the mangement of the usage example
...
 `makeindex.py <makeindex.py>`_
    Auto-generate the docs example index ``docs/source/examples.rst`` from the example scripts index ``scripts/index.rst``.
 
-`Makefile <Makefile>`_
-   A makefile allowing use of the command ``make`` to update auto-generated Jupyter notebooks. Run as ``make no-execute=true`` to update the notebooks without executing them. Use of `makejnb.py` rather than this makefile is recommended.
-
-`pytojnb.sh <pytojnb.sh>`_
-   Low-level python to Jupyter notebook conversion script. Used by both the makefile and `makejnb.py <makejnb.py>`_.
-
 `scriptcheck.sh <scriptcheck.sh>`_
```
Michael-T-McCann, line 49: This is cool, didn't know we had it.

examples/makenotebooks.py
```diff
@@ -0,0 +1,202 @@
...
+    "--no-ray",
+    action="store_true",
+    help="Execute notebooks serially, without the use of ray parallelization",
+)
+argparser.add_argument("filename", nargs="*", help="Optional Python example script filenames")
+args = argparser.parse_args()
+
+# Raise error if ray needed but not present
+if not have_ray and not args.no_ray:
+    raise RuntimeError("The ray package is required to run this script")
```
Michael-T-McCann, line 135: Is there a mechanism like: `makenotebook XXXX` that makes a single notebook without ray? I think I would use it if it existed.

Michael-T-McCann, line 135: Add ray to `docs_requirements.txt` or `examples_requirements.txt`?

Michael-T-McCann: hint: look at the help

examples/makenotebooks.py
```diff
@@ -0,0 +1,202 @@
...
+                scriptnames.append(m.group(2))
+
+# Ensure list entries are unique
+scriptnames = list(set(scriptnames))
+
+# Construct script paths
+scripts = [Path("scripts") / Path(s) for s in scriptnames]
+
+# Display status information
+print(f"Processing scripts {', '.join(scriptnames)}")
```
Michael-T-McCann, line 158: perhaps indicate somehow which scripts are actually getting built (i.e. say when one passes the check in 168)

examples/makenotebooks.py
Michael-T-McCann: hm, but it's not required with --no-ray.

Michael-T-McCann (approved):
I'm happy to see the makefile go, and I was able to use the new mechanism to build the examples on toro.

examples/makenotebooks.py
bwohlberg: I'm not entirely sure. The script can be used without `ray` by using the ``--no-ray`` flag, so it's useful, and enhances capabilities, but is not required. We've tended to be quite strict on the definition of a "requirement" (e.g. see handling of the ASTRA dependency), but I'm fine with including `ray` if it makes sense.

examples/makenotebooks.py
bwohlberg: Fixed in 53dc862. That was a bit more painful than expected!

Michael-T-McCann (approved):

PR #112, opened 2021-11-22
Modified SquareL2Loss to inherit from WeightedSquareL2Loss

FernandoDavis:
Per @tbalke 's discussion with Luke I modified the SquaredL2Loss class to inherit from WeightedSquaredL2Loss and use W=None. I tested that it ran without issue and included the documentation for __call__ from the original SquaredL2Loss into the Weighted class.


tbalke:
Regarding the decreased codecov I would suggest looking into removing the else clause in 

https://github.com/lanl/scico/blob/01a4e142e39b1a284a99164604e2628380e8be59/scico/admm.py#L209-L215

Because now `WeightedSquareL2Loss` is a super class of `SquareL2Loss` and thus
```python
isinstance(self.admm.f, WeightedSquaredL2Loss)
```
is always true.

scico/loss.py
```diff
@@ -268,6 +171,11 @@ def __init__(
             self.has_prox = True
 
     def __call__(self, x: Union[JaxArray, BlockArray]) -> float:
+        r"""Evaluate this loss at point :math:`\mb{x}`.
+
+        Args:
+            x : Point at which to evaluate loss.
+        """
```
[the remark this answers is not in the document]
FernandoDavis: That's true. I added it just for the sole reason that SquaredL2Loss had it. But considering ultimately that they all use the same base class I would like to hear Brendt's thoughts on it too.

scico/admm.py
```diff
@@ -210,9 +210,6 @@ def compute_rhs(self) -> Union[JaxArray, BlockArray]:
             if isinstance(self.admm.f, WeightedSquaredL2Loss):
```
bwohlberg, line 210: The `else` clause has already been discussed, but is there any point to this test either? The `__init__` raises an exception if `f` is not quadratic. Does that not guarantee that this test passes with the new class structure?

tbalke: Good point. I think we can remove that check.

Also, in 
https://github.com/lanl/scico/blob/a9db83098514dbca6bcdad8fc2e8e229ca37c988/scico/admm.py#L169-L177
we may rather want to test whether 
- A is instance of linop
- f is instance of WeightedSquaredL2Loss
(which implies it is quadratic)

I think it is conceptually better because all we want is to be able to evaluate
```python
         ATWy = self.admm.f.A.adj(self.admm.f.W.diagonal * self.admm.f.y) 
         rhs += 2.0 * self.admm.f.scale * ATWy 
```

Because technically we **_don't_** handle any case where
- f is not instance of WeightedSquaredL2Loss
- f is quadratic

bwohlberg: I'm inclined to agree. Is there even a good reason to keep the `is_quadratic` attribute?

tbalke: No. That part of the admm.py is the only place it is being used. I'll take care of both things.

scico/loss.py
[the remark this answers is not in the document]
bwohlberg: The docstring is the same as that from the parent class, right? I don't think we've ever discussed this or formulated an explicit policy, but my impression is that general approach adopted in the code is that the docstring of a method of a derived class is not defined unless it needs to be different to that of the parent class.

scico/admm.py
tbalke: 7455f1a

FernandoDavis: I can try and update the test case to check for the instance of weighted to get the coverage up.

tbalke: IMO adding a test for when `not(f is None or f is WeightedSquaredL2Loss)` would not be very useful.

@bwohlberg do we have a policy that we desire every branch of execution to be covered by the tests?

bwohlberg: That is best practice.

scico/admm.py
```diff
@@ -170,11 +170,14 @@ def internal_init(self, admm):
         if admm.f is not None:
             if not isinstance(admm.f.A, LinearOperator):
                 raise ValueError(
-                    "LinearSubproblemSolver requires f.A to be a scico.linop.LinearOperator; "
+                    f"LinearSubproblemSolver requires f.A to be a scico.linop.LinearOperator; "
                     f"got {type(admm.f.A)}"
```
bwohlberg, line 174: To nitpick, I don't think the `f` is necessary here.

bwohlberg, line 174: This one *is* necessary.

scico/loss.py
```diff
@@ -261,13 +161,15 @@ def __init__(
         else:
             self.is_smooth = None
 
-        if isinstance(self.A, linop.LinearOperator):
```
bwohlberg, line 264: This seems to have been a serious oversight in out discussion of removing the `is_quadratic` attribute: `A` need not be a linear operator, and if it isn't, our assumptions about the loss function are violated. Is there any way of avoiding restoring this attribute?

scico/loss.py
```diff
@@ -261,13 +161,15 @@ def __init__(
...
-
         if isinstance(self.A, linop.Diagonal) and isinstance(self.W, linop.Diagonal):
             self.has_prox = True
 
     def __call__(self, x: Union[JaxArray, BlockArray]) -> float:
+        r"""Evaluate this loss at point :math:`\mb{x}`.
+
+        Args:
+            x : Point at which to evaluate loss.
+        """
```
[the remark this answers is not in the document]
tbalke: Yes.

scico/loss.py
tbalke: > A need not be a linear operator, and if it isn't, our assumptions about the loss function are violated

Can you elaborate? It does not need to be a linop in terms of the admm?

scico/admm.py
tbalke: You are right. I added it though to avoid mixing stings and f-strings. (Similar to preferring `2.0 * np.pi` to `2 * np.pi` where the `.0` is not necessary.)

Feeling strongly about removing this?

bwohlberg: No strong feelings. I primarily mentioned it because I seemed to recall that `flake8` or `pylint` complained about it, but on checking that doesn't seem to be the case, so consider the comment withdrawn.

scico/loss.py
bwohlberg: Ignore my comment. I had forgotten about the earlier `if not isinstance(admm.f.A, LinearOperator):` test.

scico/loss.py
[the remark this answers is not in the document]
tbalke: removed in 441f756

scico/loss.py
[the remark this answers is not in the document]
tbalke: removed in 441f756

bwohlberg (approved):

PR #114, opened 2021-11-22
Stop codecov checking the patch

Michael-T-McCann:


bwohlberg:
Branched off `smajee/test-fix-astra` instead of `main`?

Michael-T-McCann:
> Branched off `smajee/test-fix-astra` instead of `main`?

fixed, thanks

PR #115, opened 2021-11-24
Add missing requirement

tbalke:
Needed in 
https://github.com/lanl/scico/blob/4395463890ad4f28416e59c1b88831f00826cded/examples/scripts/deconv_microscopy_allchn_tv_admm.py#L43-L46


bwohlberg:
If this PR is going to be accepted, then the try/except construction in the script should be replaced with a simple import.

bwohlberg (approved):

PR #117, opened 2021-11-30
New optimization algorithms

bwohlberg:
Add linearized ADMM and PDHG optimization algorithms.

See also lanl/scico-data#7.


scico/ladmm.py
```diff
@@ -0,0 +1,365 @@
+# -*- coding: utf-8 -*-
```
Michael-T-McCann, line 1: With the addition of these optimizers, I find having each optimizer as a top-level module odd. I.e., looking at the table of contents for the API reference (https://scico--117.org.readthedocs.build/en/117/_autosummary/scico.html), we have `scico.admm` and `scico.ladmm` at the same level as `scico.linop`. I propose that it's time for a `scico.optimizer` package.

scico/ladmm.py
```diff
@@ -0,0 +1,365 @@
...
+from scico.typing import JaxArray
+from scico.util import Timer, ensure_on_device
+
+__author__ = """\n""".join(
+    ["Luke Pfister <luke.pfister@gmail.com>", "Brendt Wohlberg <brendt@ieee.org>"]
+)
+
+
+class LinearizedADMM:
+    r"""Linearized Alternating Direction Method of Multipliers algorithm.
```
Michael-T-McCann, line 31: As per style guide: "Linearized alternating direction method of multipliers algorithm"

scico/ladmm.py
```diff
@@ -0,0 +1,365 @@
...
+    :class:`.LinearOperator`.
+
+    The optimization problem is solved by introducing the splitting
+    :math:`\mb{z} = C \mb{x}` and solving
+
+    .. math::
+        \argmin_{\mb{x}, \mb{z}} \; f(\mb{x}) + g(\mb{z}) \;
+       \text{such that}\; C \mb{x} = \mb{z} \;,
+
+    via a Linearized ADMM algorithm :cite:`yang-2012-linearized`
```
Michael-T-McCann, line 52: "linearized"

scico/primaldual.py
```diff
@@ -0,0 +1,284 @@
...
+from scico.linop import LinearOperator
+from scico.numpy.linalg import norm
+from scico.typing import JaxArray
+from scico.util import Timer, ensure_on_device
+
+__author__ = "Brendt Wohlberg <brendt@ieee.org>"
+
+
+class PDHG:
+    r"""Primal–Dual Hybrid Gradient algorithm.
```
Michael-T-McCann, line 29: "Primal–dual hybrid gradient algorithm"

scico/primaldual.py
```diff
@@ -0,0 +1,284 @@
...
+
+    |
+
+    Primal–Dual Hybrid Gradient (PDHG) is a family of algorithms
```
Michael-T-McCann, line 33: "Primal–dual hybrid gradient"

Michael-T-McCann (approved):
"style police" comments are FYI rather than blockers, I can tidy them up later if you wish.

scico/ladmm.py
bwohlberg: Fixed.

scico/ladmm.py
bwohlberg: Fixed.

scico/primaldual.py
bwohlberg: Fixed.

scico/primaldual.py
bwohlberg: Fxed.

scico/ladmm.py
bwohlberg: That's a good suggestion, but there are a few questions to resolve regarding the implementation:

- Naming:
  - Instead of `optimizer`, I would go with `optimize`, for consistency with `scipy.optimize`.
  - Are we going to also move `scico.solver` into this subpackage? If not, the name should perhaps reflect specific properties that differentiate it from the functions in `scico.solver`, e.g. `scico.proxopt`, since all the algorithms that would be within the subpackage would be proximal algorithms.
- Structure. Do we
  - retain the existing modules, so that e.g. `scico.admm.ADMM` becomes ` `scico.optimize.admm.ADMM`, or
  - follow the same approach as in e.g.`scico.linop`, so that `scico.admm.ADMM` becomes `scico.optimize.ADMM`.
  
My leaning is towards naming it `scico.optimize`, which would include `scico.solver`, and making the classes direct members of this module, e.g. `scico.optimize.ADMM`.

scico/ladmm.py
bwohlberg: Similar style changes still need to be applied to existing modules, such as `scico.admm`. Can you take care of that within the `mike/docs` branch?

scico/ladmm.py
bwohlberg: Discussion moved to #122.

PR #118, opened 2021-12-03
Add optional relaxation to ADMM solver

bwohlberg:
- Add optional relaxation to ADMM solver.
- Simplify ADMM step implementation.

### DIFF
--- a/scico/admm.py
+++ b/scico/admm.py
@@ -566,7 +566,7 @@ class ADMM:
         return u_list
 
     def step(self):
-        """Perform a single ADMM iteration.
+        r"""Perform a single ADMM iteration.
 
         The primary variable :math:`\mb{x}` is updated by solving the the
         optimization problem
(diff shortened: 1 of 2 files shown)


scico/admm.py
```diff
@@ -566,58 +565,49 @@ def u_init(self, x0: Union[JaxArray, BlockArray]):
...
-            \norm{\mb{z}^{(k)}_i - \mb{u}^{(k)}_i - C_i \mb{x}}_2^2
+    def step(self):
+        r"""Perform a single ADMM iteration.
 
-        """
-        return self.subproblem_solver.solve(x)
+        The primary variable :math:`\mb{x}` is updated by solving the the
+        optimization problem
 
-    def z_and_u_step(self, u_list, z_list):
```
Michael-T-McCann, line 579: I suspect the code was written this way so that the `z_and_u_step` could be jitted (jitted functions should be pure, no side effects). However, `z_and_u_step` was not actually jitted as far as I can tell. So I don't mind the refactor.

Michael-T-McCann (approved):

scico/admm.py
bwohlberg: That was my suspicion too. If we decide there is some benefit to jitting, we can refactor again, but for now this simpler structure makes more sense.

PR #126, opened 2021-12-09
Run pytest on latest jax version

### LINKED ISSUE #80: Make tests run on the latest version of jax
Thanks to #73, SCICO depends on a specific jax version which won't always be the latest one. In order to spot compatibility changes coming down the pike and to help us keep up with jax, we could make the CI run tests with both the pinned jax version and the current latest jax version.

Links from Luke:
* see here: https://docs.github.com/en/actions/learn-github-actions/managing-complex-workflows#using-a-build-matrix
* for concrete example of setting up a build matrix, see https://github.com/google/jax/blob/main/.github/workflows/ci-build.yaml

Michael-T-McCann:
This involves some code duplication in `pytest.yml`, but it gets the job done and doesn't require me (or later maintainers) to learn much about github workflow syntax. 

Closes #80.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -1,4 +1,4 @@
-# This is a basic workflow to help you get started with Actions
+# Install scico requirements and run pytest
 
 name: test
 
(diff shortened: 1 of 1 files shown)


.github/workflows/pytest.yml
```diff
@@ -49,3 +47,29 @@ jobs:
           flags: unittests
           name: codecov-umbrella
           verbose: true
+
+  pytest-latest-jax:
+    # The type of runner that the job will run on
```
bwohlberg, line 52: Perhaps a bit more maintainable to put this part in a separate file?

bwohlberg (approved):
Looks good, and quite timely considering that we already have problems with the most recent jax. Consider the comment on splitting files as an optional suggestion rather than a requirement for merge approval.

PR #128, opened 2021-12-10
Add ray.tune interface and example script

bwohlberg:
Add ray.tune interface and example script.


FernandoDavis (approved):
Everything looks good to me

PR #129, opened 2021-12-11
Move optimizers into optimize package

### LINKED ISSUE #122: Add `optimize` package
With the addition of new optimizers, I find having each optimizer as a top-level module odd. I.e., looking at the table of contents for the API reference (https://scico--117.org.readthedocs.build/en/117/_autosummary/scico.html), we have `scico.admm` and `scico.ladmm` at the same level as `scico.linop`. I propose that it's time for a `scico.optimizer` package.

_Originally posted by @Michael-T-McCann in https://github.com/lanl/scico/pull/117#discussion_r762302754_

From in person discussion: Names should be like `scico.optimize.ADMM`. `scico.solver` functions should move over to `scico.optimize`.

bwohlberg:
Move optimizers into `optimize` package. Resolves #122.

Notes:
- `solvers` not moved into `optimize` to avoid import loops
- `admm` and `pgm` submodules are treated differently from `_ladmm` and `_primaldual` because they include auxiliary classes for which the retention of the submodule prefix is desirable, e.g. `admm.LinearSubproblemSolver`
- class `ADMM` can be imported from either `scico.optimize` or `scico.optimize.admm`, and similarly for the PGM solver classes


examples/scripts/ct_astra_tv_admm.py
```diff
@@ -9,7 +9,7 @@
 =====================================
 
 This example demonstrates the use of class
-[admm.ADMM](../_autosummary/scico.admm.rst#scico.admm.ADMM) to solve a
+[admm.ADMM](../_autosummary/scico.optimize.admm.rst#scico.optimize.admm.ADMM) to solve a
```
Michael-T-McCann, line 12: Maybe I missed it in the meeting: Is it not possible to make set `ADMM = admm.ADMM` in `scico.optimize.__init__.py`? (In order to the duplication of `.admm.ADMM`.)

Michael-T-McCann: Ah, looking later on, I see
    
    from scico.optimize.admm import ADMM, LinearSubproblemSolver
    
which would become

    from scico.optimize import ADMM
    from scico.optimize.admm import LinearSubproblemSolver
    
which is maybe more confusing.

scico/optimize/pgm.py
```diff
@@ -396,21 +408,20 @@ def __init__(
         r"""
 
         Args:
-            f: Loss or Functional object with `grad` defined
-            g: Instance of Functional with defined prox method
+            f: Loss or Functional object with `grad` defined.
+            g: Instance of Functional with defined prox method..
```
Michael-T-McCann, line 412: double period

Michael-T-McCann (approved):
Despite the slight awkwardness you mentioned around `LinearSubproblemSolver`, this looks like a significant improvement in clarify.

examples/scripts/ct_astra_tv_admm.py
bwohlberg: Exactly. If one only needs `ADMM` then one can use `from scico.optimize import ADMM`, but importing from `scico.optimize.admm` is also supported to allow single-line imports.

scico/optimize/pgm.py
bwohlberg: fixed

examples/scripts/ct_astra_tv_admm.py
bwohlberg: But good thing you commented on that specific change. Many API cross-references in the examples were broken.

PR #130, opened 2021-12-12
Improve iteration statistics mechanism

bwohlberg:
- Improved iteration statistics mechanism provides option for subsampling of displayed iterations
- Modified mechanism for passing iteration statistics options in optimizer classes


scico/admm.py
```diff
@@ -416,16 +415,16 @@ def __init__(
...
-                be displayed.
-            itstat: A tuple (`fieldspec`, `insertfunc`), where `fieldspec`
-                is a dict suitable for passing to the `fields` argument
-                of the :class:`.diagnostics.IterationStats` initializer,
-                and `insertfunc` is a function with two parameters, an
-                integer and an ADMM object, responsible for constructing
-                a tuple ready for insertion into the
-                :class:`.diagnostics.IterationStats` object. If None,
-                default values are used for the tuple components.
+            itstat_options: A dict of named parameters to be passed to
```
Michael-T-McCann, line 418: Thought: for `subproblem_solver` we pass the constructed object, for `itstat_options` we pass a dict that gets used to construct an object. How are we making these decisions? Is it just that there is only one possible object that can be used with `itstat_options`?

Michael-T-McCann (approved):

scico/admm.py
bwohlberg: That's a reasonable point. As you point out, though, `subproblem_solver` can take instances of different classes, while the internal `IterationStats` object is always of one specific type, so it would seem over-complicated to pass in the initialized object rather than a specification of its parameters.

PR #131, opened 2021-12-12
Fix bugs in linop.optics

bwohlberg:
- fix bugs in `linop.optics`
- add tests for methods that were broken
- rename method `check_sampling` to `adequate_sampling`


lukepfister (approved):
lgtm!

PR #137, opened 2021-12-16
Changes to examples scripts

bwohlberg:
* Improve notebook build script
  - don't terminate entire run on single notebook execution failure
  - clean up workers after keyboard interrupt
* Minor edits to example script docstrings

### DIFF
--- a/examples/makenotebooks.py
+++ b/examples/makenotebooks.py
@@ -11,10 +11,12 @@
 import argparse
 import os
 import re
+import signal
 from pathlib import Path
 from timeit import default_timer as timer
 
 import nbformat
+import psutil
 from nbconvert.preprocessors import CellExecutionError, ExecutePreprocessor
 from py2jn.tools import py_string_to_notebook, write_notebook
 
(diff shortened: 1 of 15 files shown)


examples/makenotebooks.py
```diff
@@ -105,8 +107,10 @@ def execute_notebook(fname):
         with open(fname, "w", encoding="utf-8") as f:
             nbformat.write(nb, f)
     except CellExecutionError:
-        raise Exception(f"Error executing the notebook {fname}")
+        print(f"ERROR executing {fname}")
+        return False
```
FernandoDavis, line 111: I understand the print for the error is to avoid interrupting the execution, but is the return bool necessary? Seeing as there isn't any usage of that return within the makenotebook file.

Would it be better perhaps to have it this way?

```
except CellExecutionError:
    print(...)
    return # void return
print(...)
```

FernandoDavis (approved):
Everything looks fine to me! There's one nitpick but that shouldn't hold back the PR and can be discussed on Monday with the other pending edit.

examples/makenotebooks.py
bwohlberg: I added that so that the calling context could determine whether the execution failed. It's true that the return value is not used at the moment, but is there any harm in leaving it in the function to support possible future changes in the calling code?

FernandoDavis: Yeah definitely agree on the notation used, much better than straight up interrupting the exec. Also, I have no problems with leaving it in if its something you expect will be used in the future!

PR #140, opened 2021-12-18
Modify cg_kwargs mechanism in admm.LinearSubproblemSolver

### LINKED ISSUE #69: CG solver in admm.LinearSubproblemSolver should have default relative tolerance
The default CG solver options for `admm.LinearSubproblemSolver` include a maximum number of CG iterations, but don't override the default CG solver relative tolerance (e.g. 1e-5 for `scico.solver.cg`):

https://github.com/lanl/scico/blob/0ae09a92899c8ac1baf4a48e1d06bc15d7b0839a/scico/admm.py#L144

and this is repeated in all of the example scripts, e.g.

https://github.com/lanl/scico/blob/0ae09a92899c8ac1baf4a48e1d06bc15d7b0839a/examples/scripts/ct_astra_tv_admm.py#L76

It would be much better if the `admm.LinearSubproblemSolver` initializer also included a default relative tolerance for the CG solver (e.g. 1e-3 is often sufficient for ADMM subproblems) and if the primary mechanism for controlling number of CG iterations in the example scripts were via relative tolerance rather than maximum iterations.

bwohlberg:
Change cg_kwargs mechanism so that default values are preserved unless explicitly changed.

Resolves #69 (but, contrary to the speculation in that issue, some of the examples seem to benefit from controlling both CG tolerance and maximum iterations).

### DIFF
--- a/scico/optimize/admm.py
+++ b/scico/optimize/admm.py
@@ -166,6 +166,9 @@ class LinearSubproblemSolver(SubproblemSolver):
                 documentation for :func:`scico.solver.cg` or
                 :func:`jax.scipy.sparse.linalg.cg`,
                 including how to specify a preconditioner.
+                Default values are the same as those of
+                :func:`scico.solver.cg`, except for
+                ``"tol": 1e-4`` and ``"maxiter": 100``.
             cg_function: String indicating which CG implementation to
                 use. One of "jax" or "scico"; default "scico".  If
                 "scico", uses :func:`scico.solver.cg`. If "jax", uses
--- a/scico/optimize/admm.py
+++ b/scico/optimize/admm.py
@@ -158,25 +158,29 @@ class LinearSubproblemSolver(SubproblemSolver):
             :math:`\mb{x}` update step.
     """
 
-    def __init__(self, cg_kwargs: dict = {"maxiter": 100}, cg_function: str = "scico"):
+    def __init__(self, cg_kwargs: Optional[dict] = None, cg_function: str = "scico"):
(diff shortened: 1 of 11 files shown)


scico/optimize/admm.py
```diff
@@ -158,25 +158,29 @@ class LinearSubproblemSolver(SubproblemSolver):
...
+    def __init__(self, cg_kwargs: Optional[dict] = None, cg_function: str = "scico"):
         """Initialize a :class:`LinearSubproblemSolver` object.
 
         Args:
             cg_kwargs: Dictionary of arguments for CG solver. See
-                :func:`scico.solver.cg` or
-                :func:`jax.scipy.sparse.linalg.cg`, documentation,
+                documentation for :func:`scico.solver.cg` or
+                :func:`jax.scipy.sparse.linalg.cg`,
                 including how to specify a preconditioner.
```
tbalke, line 168: Docstring should mention defaults, especially because the defaults here differ from the defaults in solver.cg.

https://github.com/lanl/scico/blob/fe2a50e54318bc9f7c2d41325b2490991bbf8c64/scico/solver.py#L245-L255

tbalke (changes requested):
Good on my side once point of https://github.com/lanl/scico/pull/140/files#r771765711 is resolved.

scico/optimize/admm.py
bwohlberg: Good point. Done.

tbalke (approved):

Michael-T-McCann (commented):
Looks good.

PR #143, opened 2021-12-19
Miscellaneous changes

bwohlberg:
- Modify `setup.py` to improve consistency in dependency specification
- Add note to docs on pinning black version in conda
- Implement more compact iteration stats initialization in optimizer classes (except PGM)
- Additional iteration stats columns for iterative ADMM subproblem solvers
- More robust implementation of `IterationStats.__init__`
- Abbreviate ADMM primal residual column name in iteration stats
- Suppress annoying pytest DeprecationWarning regarding imp in compat

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,7 +9,7 @@ Version 0.0.2   (unreleased)
 • Additional optimization algorithms: Linearized ADMM and PDHG.
 • Move optimization algorithms into ``optimize`` subpackage.
 • Additional iteration stats columns for iterative ADMM subproblem solvers.
-• Renamed "Primal Residual" to "Prml Residual" in iteration stats.
+• Renamed "Primal Rsdl" to "Prml Rsdl" in displayed iteration stats.
 • Bump pinned `jaxlib` and `jax` versions to 0.1.70 and 0.2.19 respectively.
 
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,8 @@ Version 0.0.2   (unreleased)
 
 • Additional optimization algorithms: Linearized ADMM and PDHG.
 • Move optimization algorithms into ``optimize`` subpackage.
+• Additional iteration stats columns for iterative ADMM subproblem solvers.
+• Renamed "Primal Residual" to "Prim Residual" in iteration stats.
 • Bump pinned `jaxlib` and `jax` versions to 0.1.70 and 0.2.19 respectively.
 
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,7 +9,7 @@ Version 0.0.2   (unreleased)
 • Additional optimization algorithms: Linearized ADMM and PDHG.
 • Move optimization algorithms into ``optimize`` subpackage.
 • Additional iteration stats columns for iterative ADMM subproblem solvers.
-• Renamed "Primal Residual" to "Prim Residual" in iteration stats.
+• Renamed "Primal Residual" to "Prml Residual" in iteration stats.
 • Bump pinned `jaxlib` and `jax` versions to 0.1.70 and 0.2.19 respectively.
 
 
--- a/docs/source/contributing.rst
+++ b/docs/source/contributing.rst
@@ -82,14 +82,20 @@ Installing a Development Version
       pip install -e .  # Installs SCICO from the current directory in editable mode
 
 
-9. The SCICO project uses the `Black <https://black.readthedocs.io/en/stable/>`_
+9. The SCICO project uses the `black <https://black.readthedocs.io/en/stable/>`_
    and `isort <https://pypi.org/project/isort/>`_ code formatting utilities.
-   You should set up a `pre-commit hook <https://pre-commit.com>`_ to ensure
-   any modified code passes format check before it is committed to the development repo:
+   It is important to set up a `pre-commit hook <https://pre-commit.com>`_ to
+   ensure that any modified code passes format check before it is committed to
+   the development repo:
 
    ::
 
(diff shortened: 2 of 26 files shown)


pytest.ini
```diff
@@ -2,3 +2,4 @@
 testpaths = scico/test docs
 addopts = --doctest-glob="*rst"
 doctest_optionflags = NORMALIZE_WHITESPACE
+filterwarnings = ignore::DeprecationWarning:.*.compat
```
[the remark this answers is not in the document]
bwohlberg: This should ignore all DeprecationWarnings originating in the compat module. I guess we would try to further restrict it to a specific line in compat, but this should already be specific enough to avoid problems, don't you think?

scico/optimize/admm.py
```diff
@@ -84,6 +84,7 @@ def __init__(self, minimize_kwargs: dict = {"options": {"maxiter": 100}}):
                 :func:`scico.solver.minimize`.
         """
         self.minimize_kwargs = minimize_kwargs
+        self.info = {}
```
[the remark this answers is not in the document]
bwohlberg: Yes, it is inconsistent, but I thought this was justified by the different circumstances. In `GenericSubproblemSolver`, it has to be initialized as a dict because of the mechanism used to copy in attributes from the optimization result object. In `LinearSubproblemSolver`, on the other hand, the info is directly returned in the form of a dict, so initializing to `None` seemed reasonable. But I guess we could also switch this to an empty dict if you think the inconsistency is a significant issue.

pytest.ini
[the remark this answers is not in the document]
FernandoDavis: Ah, I wasn't aware that it would filter it encased in compat. I don't mind it this way in that case! Just was unsure if it went out of scope for a second there. Thanks for the knowledge bit

scico/optimize/admm.py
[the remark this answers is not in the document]
FernandoDavis: Something good is that this works in both cases. If we somehow want to test for:

```
if self.info:
    #do something
```

the if statement picks em up as false so its all good by me.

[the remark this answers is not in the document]
bwohlberg: OK, good.

docs/source/contributing.rst
```diff
@@ -82,15 +82,21 @@ Installing a Development Version
...
-   any modified code passes format check before it is committed to the development repo:
+   It is important to set up a `pre-commit hook <https://pre-commit.com>`_ to
+   ensure that any modified code passes format check before it is committed to
+   the development repo:
 
    ::
 
       pre-commit install  # Sets up git pre-commit hooks
 
+   It is also recommended to `pin the conda package version
```
Michael-T-McCann, line 95: The current requirement is `black>=21.6b0,<22` is that incorrect? I.e., is a version in that range not compatible with the action run on github?

If not, I don't think this problem should be affecting anyone with a normal SCICO install, because black should be `pip`-installed and conform to the version number in `dev_requirements.txt`.

Michael-T-McCann (approved):

docs/source/contributing.rst
bwohlberg: The problem is that subsequent updates (e.g. `conda update --all`) have a tendency to install a version of black outside of this specified range.

CHANGES.rst
```diff
@@ -8,6 +8,8 @@ Version 0.0.2   (unreleased)
 
 • Additional optimization algorithms: Linearized ADMM and PDHG.
 • Move optimization algorithms into ``optimize`` subpackage.
+• Additional iteration stats columns for iterative ADMM subproblem solvers.
+• Renamed "Primal Residual" to "Prim Residual" in iteration stats.
```
tbalke, line 12: Is that so? Looks like it is `Prim_Rsdl`.

I would have gone with `prml_rsdl` and `dual_rsdl` as in removing vowels from 'primal' and 'residual'.

Also, why the capitalization?

Your call, though.

scico/optimize/pgm.py
```diff
@@ -458,9 +458,9 @@ def x_step(v, L):
             itstat_fields = {
                 "Iter": "%d",
                 "Time": "%8.2e",
```
tbalke, line 460: Perhaps `"Time [s]"`?

tbalke (changes requested):

CHANGES.rst
bwohlberg: > Is that so? Looks like it is `Prim_Rsdl`.

It's `Prim Rsdl` in the display column header, and `Prim_Rsdl` when accessing the data from the returned iteration statistics.

> I would have gone with `prml_rsdl` and `dual_rsdl` as in removing vowels from 'primal' and 'residual'.

Agreed. I will change it.
 
> Also, why the capitalization?

I think capitalization looks better in the column headers. I guess the default corresponding name in the data record could be converted to lower case, but I'm on the fence as to whether it's worth the effort. How strongly do you feel about this?

scico/optimize/pgm.py
bwohlberg: I see your point, but there are complications. Let's discuss.

CHANGES.rst
tbalke: > It's Prim Rsdl in the display column header, and Prim_Rsdl when accessing the data from the returned iteration statistics.

So I think the CHANGES.rst needs to be updated then.

> I think capitalization looks better in the column headers. I guess the default corresponding name in the data record could be converted to lower case, but I'm on the fence as to whether it's worth the effort. How strongly do you feel about this?

How consistent are we about upper/lower case conventions? If we generally follow the rules we should try to do it here, too.

tbalke (approved):

PR #144, opened 2021-12-22
Allow Diagonal to use broadcasting

Michael-T-McCann:
The part I'm least happy with is that I find the resulting (broadcasted) shape of `x * y` for two BlockArrays empirically, i.e., by multiplying zeros together. I would love an `snp.broadcast_shapes` but adding it was nontrivial.

### DIFF
--- a/scico/linop/_linop.py
+++ b/scico/linop/_linop.py
@@ -167,22 +167,44 @@ def valid_adjoint(
 class Diagonal(LinearOperator):
     """Diagonal linear operator."""
 
-    def __init__(self, diagonal: JaxArray, input_dtype: Optional[DType] = None, **kwargs):
+    def __init__(
+        self,
+        diagonal: Union[Array, BlockArray],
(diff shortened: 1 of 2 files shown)


bwohlberg:
> The part I'm least happy with is that I find the resulting (broadcasted) shape of `x * y` for two BlockArrays empirically, i.e., by multiplying zeros together. I would love an `snp.broadcast_shapes` but adding it was nontrivial.

As in this line?
```
output_shape = (snp.empty(input_shape) * diagonal).shape
```

Perhaps create a low-priority issue as a reminder to come back and fix this at a later date?

bwohlberg (approved):
Approved, with recommendation to create issue as a reminder to deal with less-than-optimal aspect of this solution.

Michael-T-McCann:
> Perhaps create a low-priority issue as a reminder to come back and fix this at a later date?

See #146.

PR #151, opened 2021-12-23
Move common functions in example scripts into new module

bwohlberg:
Move some common functions in example scripts into new `scico.examples` module.

Also:
- Fix docstrings in `scico.data`
- Trivial docstring edits in `examples/scripts/ct_astra_weighted_tv_admm.py`
- For `deconv_microscopy_...` examples:
   - Changed downsampling rate to 2 (see *)
   - Downsampling rate is no longer required to be even
   - Changed path of cached EPFL BIG data to `~/.cache/scico/examples`

(*) Running on GPU may now require setting the XLA environment variables
```
export XLA_PYTHON_CLIENT_ALLOCATOR=platform
export XLA_PYTHON_CLIENT_PREALLOCATE=false
```

### DIFF
--- a/docs/source/contributing.rst
+++ b/docs/source/contributing.rst
@@ -82,8 +82,7 @@ Installing a Development Version
       pip install -e .  # Installs SCICO from the current directory in editable mode
 
 
-9. The SCICO project uses the `Black <https://black.readthedocs.io/en/stable/>`_
-   and `isort <https://pypi.org/project/isort/>`_ code formatting utilities.
+9. The SCICO project uses the `Black <https://black.readthedocs.io/en/stable/>`_, `isort <https://pypi.org/project/isort/>`_ and `pylint <https://pylint.pycqa.org/en/latest/>`_ code formatting utilities.
    You should set up a `pre-commit hook <https://pre-commit.com>`_ to ensure
    any modified code passes format check before it is committed to the development repo:
 
--- a/docs/source/contributing.rst
+++ b/docs/source/contributing.rst
@@ -82,13 +82,19 @@ Installing a Development Version
       pip install -e .  # Installs SCICO from the current directory in editable mode
 
 
-9. The SCICO project uses the `Black <https://black.readthedocs.io/en/stable/>`_, `isort <https://pypi.org/project/isort/>`_ and `pylint <https://pylint.pycqa.org/en/latest/>`_ code formatting utilities.
-   You should set up a `pre-commit hook <https://pre-commit.com>`_ to ensure
-   any modified code passes format check before it is committed to the development repo:
+9. The SCICO project uses the `black <https://black.readthedocs.io/en/stable/>`_,
+   `isort <https://pypi.org/project/isort/>`_ and `pylint <https://pylint.pycqa.org/en/latest/>`_
+   code formatting utilities. It is important to set up a `pre-commit hook <https://pre-commit.com>`_ to
+   ensure that any modified code passes format check before it is committed to the development repo:
 
    ::
 
       pre-commit install  # Sets up git pre-commit hooks
+
+   It is also recommended to `pin the conda package version
+   <https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-pkgs.html#preventing-packages-from-updating-pinning>`__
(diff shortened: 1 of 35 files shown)


docs/source/contributing.rst
```diff
@@ -179,6 +191,61 @@ A feature development workflow might look like this:
...
+The following steps show how to add new data, ``new_data.npz``, to the packaged data. We assume the SCICO repository has been cloned to ``scico/``.
+
+Note that the data is located in the scico-data submodule, which is
+symlinked to ``scico/data``.  When adding new data, both the scico and
+scico-data repositories must be updated and kept in sync.
+
+
+1. Add the ``new_data.npz`` file to the ``scico/data`` directory.
+
+2. Navigate to the ``data`` directory and add/commit the new data file:
```
tbalke, line 206: ``scico/data``

examples/README.rst
```diff
@@ -10,22 +10,87 @@ Building Notebooks
...
+Updating Notebooks in the Repo
+------------------------------
+
+The recommended procedure for rebuilding notebooks for inclusion in the ``data`` submodule is:
+
+1. Add and commit the modified script(s).
+
+2. Rebuild the notebook(s) as described above.
+
+2. Add and commit the updated notebooks following the submodule handling procedure described in the developer docs.
```
tbalke, line 79: 'notebook(s)' for consistency

I would change all 'notebook(s)' to 'notebooks' though as optional singular is obvious.

examples/scripts/ct_astra_weighted_tv_admm.py
```diff
@@ -134,7 +134,7 @@ def postprocess(x):
 
 where
 
-  $$W = \mathrm{diag}\left\{\exp( \sqrt{\mathbf{y}}) \right\}.$$
+  $$W = \mathrm{diag}\left\{\exp( \sqrt{\mathbf{y}}) \right\} \;.$$
```
tbalke, line 137: Please double check correctness of the formula.

Where does the sqrt come from?
Is there a negative sign missing?

scico/examples.py
```diff
@@ -0,0 +1,231 @@
...
+        rate: Downsampling rate.
+
+    Returns:
+        Downsampled volume.
+    """
+
+    if rate == 1:
+        return vol
+
+    if np.all([n % rate == 0 for n in vol.shape]):
```
tbalke, line 166: Interesting way of doing this. Is this the same as

https://scikit-image.org/docs/stable/api/skimage.transform.html#skimage.transform.downscale_local_mean

?

tbalke (approved):

docs/source/contributing.rst
bwohlberg: These instructions have numerous other issues, including the advice to commit directly to the ``main`` branch. Since this goes considerably beyond the intended scope of this PR, I'll create an issue as a reminder that this needs to be addressed.

examples/README.rst
bwohlberg: Agreed and fixed.

examples/scripts/ct_astra_weighted_tv_admm.py
bwohlberg: We'll have to ask @lukepfister or @Michael-T-McCann to check the equation. It's quite possible that there is a spurious sqrt left over from an earlier change to how we define the weight matrix.

scico/examples.py
bwohlberg: It's the same when the volume shape is integer divisible by `rate`. Using `skimage.transform.downscale_local_mean` would be cleaner, but that would introduce an additional dependency on `skimage`.

docs/source/contributing.rst
bwohlberg: See #152.

examples/scripts/ct_astra_weighted_tv_admm.py
bwohlberg: See #153.

scico/examples.py
tbalke: We agreed in the meeting not to add the `skimage` dependency.

bwohlberg: To be revisited if `skimage` becomes a dependency for other reasons.

PR #156, opened 2022-01-03
Remove problematic array copy operations

bwohlberg:
The copy calls in the current implementations prevent solving problems using `BlockArray` variables since `copy` is not implemented for that class. The `copy` calls should not be necessary since the relevant working variables should be `DeviceArray`, and therefore immutable, due to the `ensure_on_device` calls in initializing them.

### DIFF
--- a/scico/optimize/_ladmm.py
+++ b/scico/optimize/_ladmm.py
@@ -254,7 +254,7 @@ class LinearizedADMM:
             x0: Starting point for :math:`\mb{x}`.
         """
         z = self.C(x0)
-        z_old = z.copy()
+        z_old = z
         return z, z_old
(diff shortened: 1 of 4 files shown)


Michael-T-McCann (approved):
Does this mean these routines are not tested with BlockArrays? Perhaps such tests should be added (as a new issue).

bwohlberg:
> Does this mean these routines are not tested with BlockArrays? Perhaps such tests should be added (as a new issue).

Good point. I will fix this.

bwohlberg:
> Does this mean these routines are not tested with BlockArrays? Perhaps such tests should be added (as a new issue).

Done.

PR #157, opened 2022-01-05
Address some docs build warnings

bwohlberg:
Address some docs build warnings.

### DIFF
--- a/scico/linop/_linop.py
+++ b/scico/linop/_linop.py
@@ -169,7 +169,7 @@ class Diagonal(LinearOperator):
 
     def __init__(
         self,
-        diagonal: Union[Array, BlockArray],
+        diagonal: Union[JaxArray, BlockArray],
         input_shape: Optional[Shape] = None,
         input_dtype: Optional[DType] = None,
         **kwargs,
--- a/examples/scripts/ct_astra_tv_admm.py
+++ b/examples/scripts/ct_astra_tv_admm.py
@@ -9,7 +9,7 @@ Few-View CT (ADMM w/ Total Variation)
 =====================================
 
 This example demonstrates the use of class
-[admm.ADMM](../_autosummary/scico.optimize.html#scico.optimize.ADMM) to
+[admm.ADMM](../_autosummary/scico.optimize.rst#scico.optimize.ADMM) to
 solve a few-view CT reconstruction problem with anisotropic total
 variation (TV) regularization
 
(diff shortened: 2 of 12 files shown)


scico/linop/_linop.py
```diff
@@ -169,7 +169,7 @@ class Diagonal(LinearOperator):
 
     def __init__(
         self,
-        diagonal: Union[Array, BlockArray],
```
bwohlberg, line 172: @Michael-T-McCann: I think this was originally added by you. Please conform that the change from `Array` to `JaxArray` (which silences the warning) is indeed correct.

FernandoDavis: I think in this case, changing Array to JaxArray means that the developer using the linop has to create the jax array beforehand. If we leave it as Array, it can take a numpy array or anything and convert it inside it using the ensure_on_device method. 

Essentially, the warning means that there is a conversion to JaxArray from np.array and I'm not sure if we want to allow the linop to be created using anything other than device/block arrays. To my knowledge, it doesn't affect anything code-wise due to the ensure_on_device but it brings attention to if we want to restrict it to that type.

bwohlberg: That's a reasonable argument, but it seems that all the other LinOps use `JaxArray` rather than `Array`. In fact, although `Array` is indeed defined in `scico.typing`, from a quick scan I couldn't find a single instance of it actually being used.

Michael-T-McCann: Given that other code uses `JaxArray`, let's go with `JaxArray` here.

As a general principle, I believe we are _not_ going out of our way to be interoperable with numpy.

scico/linop/_linop.py
```diff
@@ -169,7 +169,7 @@ class Diagonal(LinearOperator):
...
+        diagonal: Union[JaxArray, BlockArray],
```
Michael-T-McCann, line 172: Perhaps this union could be defined in `scico.typing` as something like `array_like` or `vector`. Not sure how far we want to go with typing now.

Michael-T-McCann (approved):

scico/linop/_linop.py
bwohlberg: That occurred to me too. As a practical matter, though, this would probably introduce some messy circular imports since `blockarray` imports `typing`. Calling it `array_like` is a reasonable choice, but it's inconsistent with the naming scheme in `typing`. Perhaps `JaxOrBlockArray`?

bwohlberg: Agreed to return to this at some other time, with low priority.

PR #158, opened 2022-01-05
Fix explanation of weight in weighted CT example

### LINKED ISSUE #153: Equation in docstring does not match corresponding code
It appears that this docstring equation does not match what is implemented in the code:

https://github.com/lanl/scico/blob/912bb9e2762af3e6083deb62a9e53b7425c2979c/examples/scripts/ct_astra_weighted_tv_admm.py#L137

This needs to be checked and corrected if confirmed to be incorrect.

Michael-T-McCann:
Changes should appear at https://scico--158.org.readthedocs.build/en/158/examples/ct_astra_weighted_tv_admm.html

closes #153

### DIFF
--- a/examples/scripts/ct_astra_weighted_tv_admm.py
+++ b/examples/scripts/ct_astra_weighted_tv_admm.py
@@ -134,11 +134,17 @@ Set up and solve the weighted reconstruction problem
 
 where
 
-  $$W = \mathrm{diag}\left\{\exp( \sqrt{\mathbf{y}}) \right\} \;.$$
+  $$W = \mathrm{diag}\left\{ \mathrm{counts} / I_0 \right\} \;.$$
+
+The data fidelity term in this formulation follows
+:cite:`sauer-1993-local` (9)
+except for the scaling by $I_0$,
+which we use to maintain balance between
+the data and regularization terms if $I_0$ changes.
 """
 lambda_weighted = 1.14e2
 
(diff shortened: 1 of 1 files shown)


examples/scripts/ct_astra_weighted_tv_admm.py
```diff
@@ -134,11 +134,17 @@ def postprocess(x):
...
 where
 
-  $$W = \mathrm{diag}\left\{\exp( \sqrt{\mathbf{y}}) \right\} \;.$$
+  $$W = \mathrm{diag}\left\{ \mathrm{counts} / I_0 \right\} \;.$$
+
+The data fidelity term in this formulation follows
+:cite:`sauer-1993-local` (9)
+except for the scaling by $I_0$,
+which we use to maintain balance between
+the data and regularization terms if $I_0$ changes.
```
bwohlberg, line 143: This unusual approach to newlines is fine in the main docs, which are primarily viewed in rendered form, but it's perhaps undesirable here since users may wish to work directly with the source example scripts.

bwohlberg:
Why changes to so many of the notebook files?

examples/scripts/ct_astra_weighted_tv_admm.py
Michael-T-McCann: fixed

Michael-T-McCann:
> Why changes to so many of the notebook files?

Good catch, I don't recall it looking like that when I pushed, but so it is now. Fixed.

PR #162, opened 2022-01-08
Add array slicing `LinearOperator`

bwohlberg:
Add a `LinearOperator` that select part of an input array via array slicing/indexing.

Is `Slice` the best choice of name for this operator? Perhaps `Index` or `Select`?


scico/linop/_linop.py
```diff
@@ -290,3 +291,41 @@ def __init__(
 
     def _eval(self, x: JaxArray) -> JaxArray:
         return snp.sum(x, axis=self.sum_axis)
+
+
+class Slice(LinearOperator):
+    """A linear operator for slicing an array."""
```
Michael-T-McCann, line 297: or block array, I wonder? I think you could use one of these to select a block from a block array, but probably not use one to select a block and index within the block, if I'm reading the [docs](https://scico.readthedocs.io/en/latest/_autosummary/scico.blockarray.html#indexed-updating) correctly.

(not saying we should add the functionality, but maybe just address this in the docstring)

bwohlberg: I think that would be worth adding. Let's discuss details.

scico/test/linop/test_linop.py
```diff
@@ -559,3 +559,39 @@ def test_sum_bad_shapes(sumtestobj, axis):
...
+    np.s_[..., 3:],
+    np.s_[1:, :-3],
+    np.s_[1:, :, :3],
+    np.s_[1:, ..., 2:],
+]
+
+
+@pytest.mark.parametrize("slc", slice_examples)
+def test_slice_eval(slicetestobj, slc):
+    x = slicetestobj.x
```
Michael-T-McCann, line 586: Possible alternative: remove `SliceTestObj` and `slicetestobj`, replace line 586 with
`x = snp.zeros((4, 5, 6, 7), dtype=dtype)`, parametrize over `dtype`.

My eyes glaze over when I see fixtures, and what is it gaining us here? Perhaps it caches `x`?

bwohlberg: I feel the same way. The approach here is a direct copy of the tests for the `Sum` operator. No arguments on simplifying the new tests, but it would seem to make sense to simplify the ones for `Sum` at the same time?

Michael-T-McCann: Agreed. I made an issue for fixing it, let's leave it as is for this PR.

scico/linop/_linop.py
bwohlberg: The syntax is already established (first axis indexes the block), but this is not supported due to limitations of the relevant `BlockArray` code:

https://github.com/lanl/scico/blob/445412ef1c4bfb017628f13b35cf8972e51b872f/scico/blockarray.py#L799-L807

As you can see, only `int` or `Ellipsis` indexing expressions are allowed, and there is no support for concatenated indexing of block and block content. It's not clear how difficult it would be to remove the former constraint (one complication would be that indexing could return another `BlockArray` rather than a `JaxArray`), but I think it would be straightforward to add support for the latter.

bwohlberg: Added support for more general indexing of BlockArray, and documented BlockArray support by Slice.

scico/blockarray.py
```diff
@@ -115,7 +117,22 @@
...
+=========================
+
+.. _blockarray_indexing:
+
+Indexing
+--------
+
+The block index is required to either be an integer or ``...``, the former
+selecting a single block and returning it as an array (*not* a singleton
+BlockArray), and the latter returning the entire BlockArray content flattened
```
Michael-T-McCann, line 129: Just checking: did you end up removing ``...``?

Michael-T-McCann (approved):

Issue #167, opened 2022-01-11
Warnings during docs build


bwohlberg:
There are currently some rather opaque warnings during docs building
```
WARNING: Cannot resolve forward reference in type annotations of "scico.flax.ConvBNBlock": name 'Module' is not defined
WARNING: Cannot resolve forward reference in type annotations of "scico.flax.DnCNNNet": name 'Module' is not defined
WARNING: Cannot resolve forward reference in type annotations of "scico.flax.ConvBNBlock": name 'Module' is not defined
WARNING: Cannot resolve forward reference in type annotations of "scico.flax.DnCNNNet": name 'Module' is not defined
scico/scico/flax.py:docstring of flax.linen.module.Module.bind:43: WARNING: Unexpected indentation.
scico/scico/flax.py:docstring of flax.linen.module.Module.bind:43: WARNING: Unexpected indentation.
```
These are clearly related to the `scico.flax` module, but it's not clear whether the problems due to some issue in that module itself, or in the imported `flax` modules.

bwohlberg:
Switching priority to high because this seems to be killing our readthedocs builds.

bwohlberg:
The `Unexpected indentation` warning seems to have been fixed in this commit 
 google/flax@ab38a002e02a700b398d640a6a9b798c9d3e8e58, but the fix doesn't seem to have made it into a release yet.

bwohlberg:
Switched back to low priority: current readthedocs build failure seems to be related to changes in `sphinx-autodoc-typehints` between 1.12 and 1.15.

PR #174, opened 2022-01-12
Silence docs build warnings

bwohlberg:
Silence docs build warnings discussed in #167. As a side effect (the desirability of which is perhaps worthy of discussion), disables documentation of inherited methods in `scico.jax` classes. In addition, the version pinning of `sphinx-autodoc-typehints` introduced in #171 is no longer necessary (also worth discussing whether this pinning should now be removed).

### DIFF
--- a/docs/docs_requirements.txt
+++ b/docs/docs_requirements.txt
@@ -2,7 +2,7 @@
 sphinx>=3.5.2
 sphinxcontrib-napoleon
 sphinxcontrib-bibtex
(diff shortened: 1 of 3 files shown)


Michael-T-McCann (approved):
("disables documentation of inherited methods in scico.jax classes" I suspect that should read "scico.flax".)

I had a look at the effect on the docs and prefer it to what we had before.

bwohlberg:
Yes, correct, I meant `scico.flax`.

PR #175, opened 2022-01-13
Move some functions to different modules

bwohlberg:
Moved some functions from `scico.util` and `scico.math` to new module `scico.array`.

It's not clear whether it's worth keeping `scico.math`; function `safe_divide` could be moved to `scico.array`, and `rel_res` would perhaps make sense in `scico.metric`.


scico/typing.py.orig
```diff
@@ -0,0 +1,51 @@
...
+DType = Any
+"""A numpy or jax dtype."""
+
+Shape = Tuple[int, ...]
+"""A shape of a numpy or jax array."""
+
+BlockShape = Tuple[Tuple[int, ...], ...]
+"""A shape of a :class:`.BlockArray`."""
+
+<<<<<<< HEAD
```
Michael-T-McCann, line 38: leftover merge markings

scico/array.py
```diff
@@ -0,0 +1,264 @@
...
+
+import scico.blockarray
+import scico.numpy as snp
+from scico.typing import ArrayIndex, Axes, AxisIndex, JaxArray, Shape
+
+__author__ = """\n""".join(
+    [
+        "Brendt Wohlberg <brendt@ieee.org>",
+        "Luke Pfister <luke.pfister@gmail.com>",
+        "Thilo Balke <thilo.balke@gmail.com>",
```
Michael-T-McCann, line 30: I think I wrote "parse_axes". BTW, how careful are we trying to be with author lists?

scico/typing.py.orig
bwohlberg: Good catch. Not sure how this file ended up in the repo. It's gone now.

scico/array.py
bwohlberg: Sorry for the omission; now fixed. In my defense, you weren't listed in `scico.util` either. On the broader question, including this does seem to be standard, but I'm not sure what the point is, and they're a pain to maintain, so I think we should consider removing them.

scico/typing.py.orig
bwohlberg: Good catch! This file was added by mistake. It's gone now.

scico/array.py
bwohlberg: Sorry for the omission - fixed now. In my defense, you weren't listed in `scico.util` either. On the broader question, I don't really see the point, and think we should consider removing them.

bwohlberg:
> Suggestion: remove `safe_divide` entirely. It's used rarely in the code, it's a one line function, and (to my eye) the one line is more expressive the the words "safe divide".

I see some value in keeping it, if only as a record of the efficient way to implement the one line. As alternative names, perhaps `no_nan_divide`, or `nan_safe_divide`?

tbalke:
I tend to share @Michael-T-McCann's sentiment and would suggest removing `save_divide` entirely.

Unless this can be solved more generally and with a wrapper replacing function values in some region? Something like
```python
# a/b but 0.0 where b==0
c = special_case((__div__, a, b), b==0, 0.0)
```

tbalke (approved):

bwohlberg:
I'm still unconvinced on removing `safe_divide`. For now, moved it to `scico.array`, renamed to `no_nan_divide`.

lukepfister:
The `safe_divide` function could benefit from some improved documentation, explaining when/why it is necessary to use.  

The problem that it solves is differentiating something like `sin(x) / x` at zero.  it has a well defined derivative at zero, but autograd will return `nan`.   this manifests to the user as bizarro `nan`s where they "shouldn't" be, and they can be hard to track down and debug.  (see the discussion https://github.com/google/jax/issues/5039#issuecomment-735430180 and in the linked thread)

so i think this function is helpful, but discoverability is an issue.  if i was a new user and i wanted to code up a sinc function, would i know to use `safe_divide`?

Michael-T-McCann:
> The `safe_divide` function could benefit from some improved documentation, explaining when/why it is necessary to use.

Something I did not realize until reading the linked comment. This is not a matter of, e.g., `sin(x) / x` being `nan` at zero (though this is true and needs to be handled). Rather, it is a matter of the way you would expect to handle that, i.e., `jnp.where(x == 0, 1.0, jnp.sin(x) / x)`, _still having a `nan` gradient at zero!_

It is a nasty jax sharp edge. Unfortunately I don't think this function can save people from the sharp edge, consider

```python
import jax
import jax.numpy as jnp

def safe_divide(x, y):  # SCICO's safe_divide
    return jnp.where(y != 0, jnp.divide(x, jnp.where(y != 0, y, 1)), 0)

def f(x):
    return jnp.where(x == 0.0, 1.0, safe_divide(jnp.sin(x), x) + x)

print(jax.grad(f)(0.0))  # should be 1.0, is 0.0
```

Michael-T-McCann (approved):

PR #176, opened 2022-01-13
Fix typing errors and set up `mypy` workflow action

bwohlberg:
Fix typing errors and set up `mypy` workflow action

To check a single file:
```
mypy --follow-imports=skip --ignore-missing-imports <filename>
```

The typing tests are now passing, with some significant limitations, as reflected in the command
```
mypy --follow-imports=skip --ignore-missing-imports  --exclude "(numpy|test)" scico
```
Most significantly, this excludes testing of `scico.numpy`, which results in 100s of typing errors, primarily due to unrecognized functions (which are dynamically constructed) in that module. Despite these limitations, I propose that we merge this PR now so that future changes are subjected to at least some typing testing. Follow-on projects would be:
1. Figure out how to annotate a dynamically generated module so that the typing system is aware of the dynamically constructed functions.
2. Revisit the numerous lines of code marked as `# type: ignore` to determine whether the error that led to its insertion can instead be resolved by improved type annotation or use of type guards.
3. Review class structure with a view to minimization violations of the Liskov substitution principle (e.g. see issues with `prox` method).


scico/functional/_denoiser.py
```diff
@@ -56,7 +56,7 @@ def __init__(self, is_rgb: bool = False):
 
         super().__init__()
 
-    def prox(self, x: JaxArray, lam: float = 1.0, **kwargs) -> JaxArray:
+    def prox(self, x: JaxArray, lam: float = 1.0, **kwargs) -> JaxArray:  # type: ignore
```
bwohlberg, line 59: Type ignore used to suppress error `Signature of "prox" incompatible with supertype "Functional"`. Is there any way of restructuring the code to avoid this violation of the Liskov substitution principle?

scico/linop/optics.py
```diff
@@ -59,13 +59,15 @@ def radial_transverse_frequency(
         raise ValueError("Invalid input dimensions; must be 1 or 2")
 
     if np.isscalar(dx):
-        dx = (dx,) * ndim
+        dx = (dx,) * ndim  # type: ignore
```
bwohlberg, line 62: Error suppression and `assert` below needed because `mypy` does not appear to be capable of understanding the implications of the conditional here.

erik-888:
> Fix typing errors and set up `mypy` workflow action
> 
> To check a file: `mypy --follow-imports=skip --ignore-missing-imports <filename>`

scico/optimize/admm.py
```diff
@@ -154,7 +154,7 @@ class LinearSubproblemSolver(SubproblemSolver):
             :math:`\mb{x}` update step.
     """
 
-    def __init__(self, cg_kwargs: Optional[dict] = None, cg_function: str = "scico"):
+    def __init__(self, cg_kwargs: Optional[dict[str, Any]] = None, cg_function: str = "scico"):
```
FernandoDavis, line 157: In line 157 dict[str, Any] should be of type Dict of the typing module if I'm not mistaken.

scico/optimize/admm.py
```diff
@@ -481,7 +481,7 @@ def __init__(
 
         # dynamically create itstat_func; see https://stackoverflow.com/questions/24733831
         itstat_return = "return(" + ", ".join(["obj." + attr for attr in itstat_attrib]) + ")"
-        scope = {}
+        scope: dict[str, Callable] = {}
```
FernandoDavis, line 484: Same as above. In line 484 dict[str, Any] should be of type Dict of the typing module.

FernandoDavis (changes requested):

scico/optimize/admm.py
Michael-T-McCann: Check out https://mypy.readthedocs.io/en/stable/builtin_types.html, looks like @FernandoDavis is correct, `dict[stuff]` is a syntax error in Python 3.8.

bwohlberg: But it's supported natively in Python 3.9, and from Python 3.7 onward when using `from __future__ import annotations`, so I think this usage is fine (and note the absence of errors in the CI).

scico/optimize/admm.py
bwohlberg: As above.

scico/functional/_denoiser.py
Michael-T-McCann: A proposal: `Union[JaxArray, BlockArray]` is used whenever a vector is expected. If the operator/functional can't deal with inputs of a certain shape (including not dealing with BlockArrays of any shape), that's a runtime error. 

By analogy, a function may take a float and then throw a runtime error if it is negative. We don't try to change the type annotation to `positive_float`.

Michael-T-McCann: Or, we do something at the subclass level, maybe making `prox` and `eval` abstract. That helps with Liskov substitution because a generic `Functional` would have no useful behavior at all.

scico/functional/_functional.py
```diff
@@ -62,10 +62,8 @@ def __call__(self, x: Union[JaxArray, BlockArray]) -> float:
            x: Point at which to evaluate this functional.
 
         """
-        if not self.has_eval:
-            raise NotImplementedError(
-                f"Functional {type(self)} cannot be evaluated; has_eval={self.has_eval}"
-            )
+        # Functionals that can be evaluated should override this method.
+        raise NotImplementedError(f"Functional {type(self)} cannot be evaluated.")
```
Michael-T-McCann, line 66: nice!

scico/functional/_denoiser.py
bwohlberg: IIRC, our code has quite a few Liskov violations, so that would be worth considering. On "A proposal", that's a reasonable analogy, but "mypy" doesn't see it that way: the type narrowing was introduced to address other typing complaints.

scico/linop/optics.py
Michael-T-McCann: Possibly this could be achieved via a type guarding version of `isscalar` https://mypy.readthedocs.io/en/latest/type_narrowing.html#user-defined-type-guards

Michael-T-McCann (approved):

scico/typing.py
```diff
@@ -9,6 +9,12 @@
 
 from typing import Any, Tuple, Union
 
+try:
+    # available in python 3.10
+    from typing import EllipsisType  # type: ignore
+except ImportError:
+    EllipsisType = Any
```
bwohlberg, line 16: @Michael-T-McCann: why `EllipsisType = Any` rather than `EllipsisType = type(Ellipsis)`?

Michael-T-McCann: No idea; I don't think I wrote this code.

scico/linop/_stack.py
```diff
@@ -35,59 +36,69 @@ def __init__(
...
 
-        self.collapse = collapse
-        output_shape = tuple(op.shape[0] for op in ops)  # assumes BlockArray output
-
-        # check if collapsable and adjust output_shape if needed
-        self.collapsable = all(output_shape[0] == s for s in output_shape)
-        if self.collapsable and self.collapse:
-            output_shape = (len(ops),) + output_shape[0]  # collapse to DeviceArray
+        if any([is_nested(op.shape[0]) for op in ops]):
+            raise ValueError("Cannot stack `LinearOperators`s with nested output shapes.")
```
FernandoDavis, line 90: LinearOperators has an extra s here.

bwohlberg:
Remaining issues to be dealt with notes as #284.

Issue #163, opened 2022-01-10
svmbir center offset


smajee:
The scico svmbir interface doesn't include the center_offset option currently. I think we should add it. It is necessary for working with real experimental data.

We could also add a kwargs in the init function to optionally pass other arguments to svmbir.

smajee:
Fixed in PR #177

PR #177, opened 2022-01-16
add center_offset to svmbir interface

smajee:
Fix issue #163

### DIFF
--- a/scico/linop/radon_svmbir.py
+++ b/scico/linop/radon_svmbir.py
@@ -57,7 +57,7 @@ class ParallelBeamProjector(LinearOperator):
             angles: Array of projection angles in radians, should be
                 increasing.
             num_channels: Number of pixels in the sinogram.
-            center_offset: Position of the image center relative to the center-of-rotation, in units of pixels
+            center_offset: Position of the detector center relative to the center-of-rotation, in units of pixels
             is_masked: If True, the valid region of the image is
                 determined by a mask defined as the circle inscribed
                 within the image boundary. Otherwise, the whole image
--- a/scico/linop/radon_svmbir.py
+++ b/scico/linop/radon_svmbir.py
@@ -57,7 +57,7 @@ class ParallelBeamProjector(LinearOperator):
             angles: Array of projection angles in radians, should be
                 increasing.
             num_channels: Number of pixels in the sinogram.
-            center_offset: Position of the image center relative to the center-of-rotation in units of (fractional) pixels
+            center_offset: Position of the image center relative to the center-of-rotation, in units of pixels
             is_masked: If True, the valid region of the image is
                 determined by a mask defined as the circle inscribed
                 within the image boundary. Otherwise, the whole image
--- a/scico/linop/radon_svmbir.py
+++ b/scico/linop/radon_svmbir.py
@@ -48,6 +48,7 @@ class ParallelBeamProjector(LinearOperator):
         input_shape: Shape,
         angles: np.ndarray,
         num_channels: int,
+        center_offset: Optional[float] = 0.0,
         is_masked: Optional[bool] = False,
     ):
         """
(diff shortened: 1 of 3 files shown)


scico/linop/radon_svmbir.py
```diff
@@ -56,13 +57,15 @@ def __init__(
             angles: Array of projection angles in radians, should be
                 increasing.
             num_channels: Number of pixels in the sinogram.
+            center_offset: Position of the image center relative to the center-of-rotation in units of (fractional) pixels
```
bwohlberg, line 60: Nitpick: ... relative to the center-of-rotation, in units ...

scico/test/linop/test_radon_svmbir.py
```diff
@@ -25,6 +25,9 @@
 BIG_INPUT = (128, 129, 200, 201)
 SMALL_INPUT = (4, 5, 7, 8)
 
+BIG_INPUT_OFFSET_RANGE = (0, 0.3, 3)
+SMALL_INPUT_OFFSET_RANGE = (0, 0.01, 0.1)
+
```
bwohlberg, line 30: The use of all upper-case variable names is consistent with prior usage in this file, but is inconsistent with scico coding standards.

bwohlberg (approved):
Approved, modulo minor comments.

scico/linop/radon_svmbir.py
smajee: Made the change

scico/test/linop/test_radon_svmbir.py
smajee: Fixed it

bwohlberg: Sorry, my mistake. I don't think this usage occurs anywhere else in scico, but @Michael-T-McCann is correct.

scico/linop/radon_svmbir.py
```diff
@@ -56,13 +57,15 @@ def __init__(
             angles: Array of projection angles in radians, should be
                 increasing.
             num_channels: Number of pixels in the sinogram.
+            center_offset: Position of the image center relative to the center-of-rotation, in units of pixels
```
Michael-T-McCann, line 60: Why is this a scalar? As described, I would expect it to be a length-2 vector, because the image is 2D.

Michael-T-McCann (changes requested):
I believe the docstring needs to be clarified, see my comment above. Further, the corresponding parameter in the [svmbir docs](https://svmbir.readthedocs.io/en/latest/svmbir.html#svmbir.project) talks about channels, here we have pixels. Different words for the same thing?

scico/test/linop/test_radon_svmbir.py
smajee: In `test_radon_astra.py`, global variables such as `rtol_cpu` are in llower case. Should those be changed to uper case as well then?

scico/linop/radon_svmbir.py
smajee: I think only a scalar needs to be inputted since an offset parallel to the x-ray doesn't change the projector, only offset perpendicular to the ray does.

smajee: (Response to the comment below, can't seem to reply there)
In the svmbir package, channel-size of the detector and pixel-size can be different but in the scico-svmbir interface, we simplify the interface to make them the same. 
Pixels seem simpler to me since `num_channels` is defined in terms of pixels in the doc-string anyway. But I am open to changing the docstring wording to channels

smajee: How about I change the parameter description to "Position of the image center relative to the center-of-rotation along the direction perpendicular to the x-rays, in units of pixels"

Michael-T-McCann: Hm, seems like my complaint is more with svmbir than with your PR. I guess for the PR, so long as it is clear which parameter of svmbir the user is setting (and I think it's already clear), I think it's fine.

Regarding the parameter, your description makes more sense. Maybe it would be even more natural to speak about the offset in the projection domain (maybe you were already trying to do that? I'm reading "image" as object-domain). This would be consistent with the use of "channels" in the svmbir docs. You also get the "perpendicular to the X-rays" part for free.

smajee: That is a nice idea. Implemented the change in commit ef0a4a4.

Michael-T-McCann (approved):

PR #178, opened 2022-01-17
Update conda install scripts

bwohlberg:
Update conda install scripts:
* Update `jaxlib` version
* Update `astra-toolbox` install method
* Default Python version now 3.9
* Handle additional dependencies for `svmbir` and `bm3d`

### DIFF
--- a/examples/notebooks_requirements.txt
+++ b/examples/notebooks_requirements.txt
@@ -4,4 +4,4 @@ nb_conda_kernels
 psutil
 py2jn
 pypandoc
-ray
+ray[tune]
--- a/misc/conda/install_conda.sh
+++ b/misc/conda/install_conda.sh
@@ -1,4 +1,4 @@
-#! /bin/bash
+#!/usr/bin/env bash
 
 # This script installs miniconda3 in the specified path
 #
--- a/misc/conda/make_conda_env.sh
+++ b/misc/conda/make_conda_env.sh
@@ -234,6 +234,14 @@ fi
 # Install other packages that require installation via pip
 pip install $NOCONDA
 
+# Warn if libopenblas-dev not installed on debian/ubuntu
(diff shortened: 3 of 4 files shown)


Michael-T-McCann (changes requested):
In my shell (zsh on mac), `./make_conda_env.sh` gives the error 

    ./make_conda_env.sh: line 63: syntax error near unexpected token `&'

bwohlberg:
Almost certainly due to incompatibilities between `zsh` and `bash`. How did you run it? It's intended as a `bash` script, and indicated as such by the initial `#! /bin/bash`.

Michael-T-McCann:
> How did you run it?

The same error occurs when I run it with `bash make_conda_env.sh` or `./make_conda_env.sh`. I believe my bash version is 3.2.

bwohlberg:
> The same error occurs when I run it with `bash make_conda_env.sh` or `./make_conda_env.sh`. I believe my bash version is 3.2.

Indeed, not incompatibility between `zsh` and `bash`, but use of an incredibly old version of bash (4.0 was released in 2009) that does not support the `case` fall through syntax using `;;&`. I added a test for bash 4 to the script.

Michael-T-McCann (commented):
I can confirm the script now leads me to update bash. After doing so:

The environment I get fails a bunch of pytests due to `scico.random`, which in turn (I think) is due to the environment giving me jax version 0.2.27 instead of the pinned version 0.2.26.

Why doesn't the workflow catch these pytest failures on 0.2.27? Because of a bug in the workflow, see #189.

Michael-T-McCann (approved):

bwohlberg:
> The environment I get fails a bunch of pytests due to `scico.random`, which in turn (I think) is due to the environment giving me jax version 0.2.27 instead of the pinned version 0.2.26.

Fixed and tested on OSX with Python 3.9.

PR #180, opened 2022-01-18
Make prox notation consistent

### LINKED ISSUE #75: Improve consistency of notation for proxes
In various places in the code and docs prox functions are written with an input of `x` and minimization over `v`, e.g.,  https://github.com/lanl/scico/blob/d01c90c523d97ddf55e81ed4e2116121828ad0c5/scico/functional/_functional.py#L78 In other places, it is the opposite (e.g., https://scico.readthedocs.io/en/latest/functional.html).

Unifying these notations would be cleaner and less confusing. Let's use the input `v`, minimization over `x` convention, which matches wikipedia, Boyd's "Proximal Algorithms", etc.

Michael-T-McCann:
Closes #75 .

### DIFF
--- a/docs/source/functional.rst
+++ b/docs/source/functional.rst
@@ -46,7 +46,7 @@ An instance of :class:`.Functional`, ``f``, may provide three core operations.
 * Proximal operator
    - ``f.prox(v, lam)`` returns the result of the scaled proximal
      operator of ``f``, i.e., the proximal operator of ``lambda x:
-     lam * f(x)``, evaulated at the point ``v``.
+     lam * f(x)``, evaluated at the point ``v``.
    - The proximal operator of a functional :math:`f : \mathbb{R}^n \to
      \mathbb{R}` is the mapping :math:`\mathrm{prox}_f : \mathbb{R}^n
      \to \mathbb{R}^n` defined as
--- a/docs/source/functional.rst
+++ b/docs/source/functional.rst
@@ -44,14 +44,16 @@ An instance of :class:`.Functional`, ``f``, may provide three core operations.
      All that is required is that the functional can be evaluated, ``f.has_eval == True``.
      However, the result may not be a valid gradient (or subgradient) for all inputs.
 * Proximal operator
-   - ``f.prox(v, lam)`` returns the result of the scaled proximal operator
-     at ``v`` with scale ``lam``.
-   - The proximal operator of a functional :math:`f : \mathbb{R}^n \to \mathbb{R}` is the mapping
-     :math:`\mathrm{prox}_f : \mathbb{R}^n \times \mathbb{R} \to \mathbb{R}^n` defined as
+   - ``f.prox(v, lam)`` returns the result of the scaled proximal
+     operator of ``f``, i.e., the proximal operator of ``lambda x:
+     lam * f(x)`` evaulated at the point ``v``.
+   - The proximal operator of a functional :math:`f : \mathbb{R}^n \to
+     \mathbb{R}` is the mapping :math:`\mathrm{prox}_f : \mathbb{R}^n
+     \to \mathbb{R}^n` defined as
 
      .. math::
-
-      \mathrm{prox}_f (v, \lambda) =  \argmin_x \lambda f(x) + \frac{1}{2} \norm{v - x}_2^2.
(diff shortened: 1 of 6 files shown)


docs/source/functional.rst
```diff
@@ -44,14 +44,16 @@ An instance of :class:`.Functional`, ``f``, may provide three core operations.
      All that is required is that the functional can be evaluated, ``f.has_eval == True``.
      However, the result may not be a valid gradient (or subgradient) for all inputs.
 * Proximal operator
-   - ``f.prox(v, lam)`` returns the result of the scaled proximal operator
-     at ``v`` with scale ``lam``.
-   - The proximal operator of a functional :math:`f : \mathbb{R}^n \to \mathbb{R}` is the mapping
-     :math:`\mathrm{prox}_f : \mathbb{R}^n \times \mathbb{R} \to \mathbb{R}^n` defined as
+   - ``f.prox(v, lam)`` returns the result of the scaled proximal
+     operator of ``f``, i.e., the proximal operator of ``lambda x:
+     lam * f(x)``, evaulated at the point ``v``.
```
bwohlberg, line 49: Typo: evaulated

scico/functional/_functional.py
```diff
@@ -76,27 +76,27 @@ def __call__(self, x: Union[JaxArray, BlockArray]) -> float:
...
-           \mathrm{prox}_{\lambda f}(\mb{x}) = \argmin_{\mb{v}}
-           \frac{1}{2} \norm{\mb{x} - \mb{v}}_2^2 + \lambda
-           \ \mathrm{f}(\mb{v}) \;,
+           \mathrm{prox}_{\lambda f}(\mb{v}) = \argmin_{\mb{x}}
+           \lambda f(\mb{x}) +
+           \frac{1}{2} \norm{\mb{v} - \mb{x}}_2^2\;,
 
-        where :math:`f(\mb{v})` represents this functional evaluated at
-        :math:`\mb{v}`.
+        where :math:`\lambda f(\mb{x})` represents this functional evaluated at
```
bwohlberg, line 92: Why the change v --> x here?

scico/functional/_norm.py
```diff
@@ -178,13 +177,17 @@ def prox(
             \end{cases}
 
         Args:
-            x:  Input array :math:`\mb{x}`.
+            v:  Input array :math:`\mb{v}`.
             lam: Proximal parameter :math:`\lambda`.
+            kwargs: Additional arguments that may be used by derived
+                classes. These include ``x0``, an initial guess for the
+                minimizer.
```
bwohlberg, line 184: Perhaps worth providing a bit more details on "the minimizer"?

Michael-T-McCann: Good point, `x0` shouldn't be mentioned at all in many of these cases. In fact, `kwargs` is not often used, but I'm not sure how better to document that.

scico/functional/_functional.py
```diff
@@ -97,7 +97,7 @@ def prox(
             lam: Proximal parameter :math:`\lambda`.
             kwargs: Additional arguments that may be used by derived
                 classes. These include ``x0``, an initial guess for the
-                minimizer.
+                minimizer in the defintion of :math:`\mathrm{prox}`.
```
bwohlberg, line 100: Typo: defintion

scico/functional/_indicator.py
```diff
@@ -51,7 +51,7 @@ def prox(
         self, v: Union[JaxArray, BlockArray], lam: float = 1.0, **kwargs
     ) -> Union[JaxArray, BlockArray]:
         r"""Evaluate the scaled proximal operator of the indicator over
-            the non-negative orthant:
+            the non-negative orthant, :math:`I`,:
```
bwohlberg, line 54: On its own, perhaps "I" might be confused for the identity matrix? Perhaps write something like I_{[\mb{x} > 0]} to explicitly include the set of which it is the indicator?

bwohlberg (approved):
Approved modulo minor additional comments.

scico/functional/_functional.py
Michael-T-McCann: Fixed and flyspell-prog-mode added to init.el

Issue #89, opened 2021-11-12
Loss is_smooth attribute 


tbalke:
Currently PGM requires the losses to be smooth (f.is_smooth = True) in order to run although at the same time some of our losses are not smooth, for example, PoissonLoss.

It is still often possible to run PGM (see Poisson examples and TV example) even though the losses are not smooth everywhere by forcing the is_smooth flag to be true.

I suggest re-thinking the use of is_smooth in conjunction with PGM.

_Originally posted by @tbalke in https://github.com/lanl/scico/pull/78#discussion_r747135635_

tbalke:
Agreement during recent meeting, to just remove the attribute.

PR #184, opened 2022-01-19
remove is_smooth attribute (#89)

tbalke:
- [ ] ~~requires~~ *consider* re-building the examples


docs/source/functional.rst
```diff
@@ -39,8 +39,7 @@ An instance of :class:`.Functional`, ``f``, may provide three core operations.
    - ``f.grad(x)`` returns the gradient of the functional evaluated at ``x``.
    - Gradients are calculated using JAX reverse-mode automatic differentiation,
      exposed through :func:`scico.grad`.
-   - A functional that is smooth has the attribute ``f.is_smooth == True``.
-   - NOTE:  The gradient of a functional ``f`` can be evaluated even if ``f.is_smooth == False``.
+   - NOTE:  The gradient of a functional ``f`` can be evaluated even if that functional is not smooth.
```
bwohlberg, line 42: Can we refer to part of the jax docs to expand on this? Also, considering using markup (e.g. *Note*) rather than all caps for emphasis.

scico/optimize/pgm.py
```diff
@@ -387,8 +387,7 @@ class PGM:
 
     Minimize a function of the form :math:`f(\mb{x}) + g(\mb{x})`.
 
-    The function :math:`f` must be smooth and :math:`g` must have a
-    defined prox.
+    The function :math:`g` must have a defined prox.
```
bwohlberg, line 390: It's worth mentioning that f should be smooth, but that the algorithm will work for f with points at which it's non-smooth as long as those points are not visited during the optimization.

bwohlberg (approved):
Approved, modulo comments.

bwohlberg:
>  requires re-building the examples

In principle this is true, but we've been a bit slack about this recently (a better policy is probably to just do this prior to a new release, or after really major changes), so I think we can let this slide for this PR.

docs/source/functional.rst
Michael-T-McCann: I had a look this morning and could not find any clear discussion of this in the jax docs. I can at least confirm that the statement is sometimes true, e.g., `jax.grad(lambda x: jnp.abs(x))(0.0) == 1.0`.

scico/optimize/pgm.py
Michael-T-McCann: Will at also work if valid subgradients are returned for nonsmooth points? It would be valuable to say so (with a ref) if it is true.

docs/source/functional.rst
tbalke: > the statement is sometimes true, e.g., `jax.grad(lambda x: jnp.abs(x))(0.0) == 1.0`

I think this is an easier example because x is defined everywhere else but at 0.

Where it breaks is e.g. at `x / (1 + jnp.sqrt(x))` which `grad`s to `nan` at zero. [See](https://jax.readthedocs.io/en/latest/notebooks/Custom_derivative_rules_for_Python_code.html?highlight=differentiable#enforcing-a-differentiation-convention)

I don't know what the general rule for scico is, though.

tbalke: > Also, considering using markup

See 0fdfcf6

scico/optimize/pgm.py
tbalke: Perhaps @crstngc can comment.

tbalke: > It's worth mentioning that f should be smooth, but that the algorithm will work for f with points at which it's non-smooth as long as those points are not visited during the optimization.

Generally agree, however,

> as long as those points are not visited

Does *as long as* mean the same as *if and only if*?

tbalke: Let's say

_guaranteed to work if globally smooth_

(leaving it open to the interpretation of the user what happens with not globally smooth losses)

tbalke: See 02fe9bb

Michael-T-McCann (approved):

scico/optimize/pgm.py
```diff
@@ -387,7 +387,8 @@ class PGM:
 
     Minimize a function of the form :math:`f(\mb{x}) + g(\mb{x})`.
 
-    The function :math:`g` must have a defined prox.
+    The function :math:`g` must have a defined prox and convergence is
+    guaranteed if :math:`f` is smooth.
```
bwohlberg, line 391: This isn't really true since `f` and `g` are usually also assumed to be convex. It's worth noting that the `ADMM` docs simply specify what the required types of these arguments are, without getting into convergence requirements. For now, perhaps better just to follow the `ADMM` example, i.e.
```
Solve an optimization problem of the form

    .. math::
        \argmin_{\mb{x}} \; f(\mb{x}) + g(\mb{x}) \;,

    where :math:`f` and :math:`g` are instances of :class:`.Functional`.
```
and also open a new issue as a reminder that more careful specifications (such as when functionals need to have proximal operators defined) should be added to all the optimizer classes.

scico/optimize/pgm.py
```diff
@@ -552,7 +553,8 @@ class AcceleratedPGM(PGM):
 
     Minimize a function of the form :math:`f(\mb{x}) + g(\mb{x})`.
 
-    The function :math:`g` must have a defined prox. The accelerated
+    The function :math:`g` must have a defined prox and convergence is
+    guaranteed if :math:`f` is smooth. The accelerated
```
bwohlberg, line 557: See comment above.

scico/optimize/pgm.py
tbalke: see 5af596b

scico/optimize/pgm.py
tbalke: see 5af596b

Issue #93, opened 2021-11-15
Output types of operators


tbalke:
The question got raised in https://github.com/lanl/scico/pull/88#discussion_r748994235 since the svmbir and BM3D routines returned ndarrays which may be a problematic type to pop up inside the solvers.

In general this poses the questions:
- what types do our routines return?
- do we want consistency (as in `type_in=type_out` or `type_out=Device-array-like`)

lukepfister:
You almost certainly want `type_out=Device-array-like` in any public interfaces.  Avoid plain ndarrays as much as possible;  they will break autograd/jit/etc if they leak out.

Michael-T-McCann:
Narrow version of the issue, change svmbir and BM3D to return jax arrays.

PR #187, opened 2022-01-24
Make output types jax-array

smajee:
Fixes output types of svmbir and bm3d to be always jaxarrays. Fixes issue #93

### DIFF
--- a/scico/functional/_denoiser.py
+++ b/scico/functional/_denoiser.py
@@ -11,7 +11,6 @@ from typing import Optional
 
 import numpy as np
 
-import jax
 from jax.experimental import host_callback as hcb
 
 from bm3d import bm3d, bm3d_rgb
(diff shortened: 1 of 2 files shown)


scico/functional/_denoiser.py
```diff
@@ -110,7 +111,7 @@ def prox(self, x: JaxArray, lam: float = 1.0, **kwargs) -> JaxArray:
         # undo squeezing, if neccessary
         y = y.reshape(x_in_shape)
 
-        return y
+        return jax.device_put(y)
```
bwohlberg, line 114: The return type here was a numpy ndarray? I would have expected `hcb.call` to keep things within the jax world.

smajee: You are right, the device_put wasn't necessary here. Fixed it.

Issue #132, opened 2021-12-13
Use same defaults handling in svmbir as in iteration stats


tbalke:
Refer to #130

Dictionary with missing arguments should result in defaults of high level code.

bwohlberg:
See also #69 for a related issue.

Issue #95, opened 2021-11-15
Positivity of svmbir prox


tbalke:
Currently, it is the case that when `positivity=True` in our `SVMBIRWeightedSquaredL2Loss`, the prox is a proper prox, but not of the loss of the `SVMBIRWeightedSquaredL2Loss` but of the sum of the `SVMBIRWeightedSquaredL2Loss` and an non-zero indicator.

- [x] First this should be properly documented in the code. (Addressed in 079ac16)

- [ ] Secondly, since the positivity flag may be useful, we should keep it but adjust the loss eval so that it is correct.

- [ ] Lastly, we should create some tests to check whether the corrected loss yields the same prox as the internal svmbir prox.

smajee:
We discussed today about having a separate svmbir loss (say `SVMBIRExtendedLoss`) which allows setting the positivity and masking as directed by the user. The `SVMBIRWeightedSquaredL2Loss` loss would then implement no masking and no positivity so as to be a proper `WeightedSquaredL2Loss`.

It feels to me that there might be a lot of code/docs duplication between these two classes with some small differences. Any ideas on how to reduce the code duplication? One (albeit suboptimal) approach could be to make `SVMBIRWeightedSquaredL2Loss` inherit from `SVMBIRExtendedLoss`.

tbalke:
I agree that the 'is-a' relationship is not very clean through inheritance.

**_But if the only concern_** is code duplication we could perhaps define generic functions outside the classes. Something like:

```python

def generic_prox(self, v, lam, positivity=None):
    pass # generic definition here

class SVMBIRExtendedLoss:
    ...
    def prox(self, v, lam):
        return generic_prox(self, v, lam, positivity=self.positivity)
    ...

class SVMBIRWeightedSquaredL2Loss(WeightedSquaredL2Loss):
    ...
    def prox(self, v, lam):
        return generic_prox(self, v, lam, positivity=False)
    ...

**Comment by bwohlberg** [comment]
That's a reasonable option, but I think deriving from a common private base class achieves the same thing in a slightly cleaner way:
```python

class _SVMBIRCommon:
    ...
    def prox(self, v, lam):
       pass # generic definition here

class SVMBIRExtendedLoss(Loss, _SVMBIRCommon):
    ...
    # no need for prox re-definition
    ...

class SVMBIRWeightedSquaredL2Loss(WeightedSquaredL2Loss, _SVMBIRCommon):
        ...
    # no need for prox re-definition
    ...
```

[...]

PR #191, opened 2022-01-27
Refactor svmbir loss

smajee:
Fix issue #95 
Fix issue #132

### DIFF
--- a/scico/linop/radon_svmbir.py
+++ b/scico/linop/radon_svmbir.py
@@ -204,7 +204,7 @@ class SVMBIRExtendedLoss(Loss):
     def __init__(
         self,
         *args,
-        prox_kwargs: dict = {},
+        prox_kwargs: Optional[dict] = None,
         positivity: bool = False,
         W: Optional[Diagonal] = None,
         **kwargs,
@@ -230,10 +230,13 @@ class SVMBIRExtendedLoss(Loss):
 
         self.has_prox = True
 
+        if prox_kwargs is None:
+            prox_kwargs = {}
+
         default_prox_args = {"maxiter": 1000, "ctol": 0.001}
         default_prox_args.update(prox_kwargs)
 
-        svmbir_prox_args = dict()
+        svmbir_prox_args = {}
         if "maxiter" in default_prox_args:
             svmbir_prox_args["max_iterations"] = default_prox_args["maxiter"]
         if "ctol" in default_prox_args:
@@ -311,7 +314,7 @@ class SVMBIRWeightedSquaredL2Loss(SVMBIRExtendedLoss, WeightedSquaredL2Loss):
     def __init__(
         self,
         *args,
-        prox_kwargs: dict = {},
+        prox_kwargs: Optional[dict] = None,
         **kwargs,
     ):
         r"""Initialize a :class:`SVMBIRWeightedSquaredL2Loss` object.
--- a/scico/linop/radon_svmbir.py
+++ b/scico/linop/radon_svmbir.py
@@ -109,13 +109,15 @@ class ParallelBeamProjector(LinearOperator):
         center_offset: Optional[float] = 0.0,
         roi_radius: Optional[float] = None,
     ) -> JaxArray:
-        return svmbir.project(
-            np.array(x),
-            np.array(angles),
-            num_channels,
-            verbose=0,
-            center_offset=center_offset,
-            roi_radius=roi_radius,
+        return jax.device_put(
+            svmbir.project(
+                np.array(x),
+                np.array(angles),
+                num_channels,
+                verbose=0,
+                center_offset=center_offset,
+                roi_radius=roi_radius,
+            )
         )
 
     def _proj_hcb(self, x):
@@ -143,14 +145,16 @@ class ParallelBeamProjector(LinearOperator):
         center_offset: Optional[float] = 0.0,
         roi_radius: Optional[float] = None,
     ):
-        return svmbir.backproject(
-            np.array(y),
-            np.array(angles),
-            num_rows,
-            num_cols,
-            verbose=0,
-            center_offset=center_offset,
-            roi_radius=roi_radius,
+        return jax.device_put(
+            svmbir.backproject(
+                np.array(y),
+                np.array(angles),
+                num_rows,
+                num_cols,
+                verbose=0,
+                center_offset=center_offset,
+                roi_radius=roi_radius,
+            )
         )
 
     def _bproj_hcb(self, y):
--- a/scico/linop/radon_svmbir.py
+++ b/scico/linop/radon_svmbir.py
@@ -193,14 +193,10 @@ class SVMBIRExtendedLoss(Loss):
     :class:`scico.linop.Identity`.
 
     The extended loss differs from a typical weighted squared :math:`\ell_2` loss
-    is the following ways.
-
+    is the following aspects.
     When ``positivity=True``, the prox projects onto the non-negative
-    orthant, but the the loss, :math:`\alpha l(\mb{y}, A(\mb{x}))`,
-    is unaffected by this setting and still evaluates to finite values
-    when :math:`\mb{x}` is not in the non-negative orthant.
-
-    When the ``is_masked`` option of the associated :class:`.ParallelBeamProjector` is True,
+    orthant and the loss is infinite if any element of the input is negative.
+    When the ``is_masked`` option of the associated :class:`.ParallelBeamProjector` is `True`,
     the reconstruction is computed over a masked region of the image as described
     in class :class:`.ParallelBeamProjector`.
     """
--- a/scico/linop/radon_svmbir.py
+++ b/scico/linop/radon_svmbir.py
@@ -228,7 +228,7 @@ class SVMBIRExtendedLoss(Loss):
                 causes the defaults of the underlying :meth:`svmbir.recon`
                 prox routine to be used.
             positivity: Enforce positivity in the prox operation. The
-                loss is not affected.
+                loss is infinity if any element of the input is negative.
         """
         super().__init__(*args, **kwargs)
 
@@ -258,6 +258,13 @@ class SVMBIRExtendedLoss(Loss):
                 raise Exception(f"The weights, W.diagonal, must be non-negative.")
         else:
             raise TypeError(f"W must be None or a linop.Diagonal, got {type(W)}")
+
+    def __call__(self, x: JaxArray) -> float:
+
+        if self.positivity and snp.sum(x < 0) > 0:
+            return snp.inf
+        else:
+            return self.scale * (self.W.diagonal * snp.abs(self.y - self.A(x)) ** 2).sum()
 
     def prox(self, v: JaxArray, lam: float, **kwargs) -> JaxArray:
         v = v.reshape(self.A.svmbir_input_shape)
--- a/scico/linop/radon_svmbir.py
+++ b/scico/linop/radon_svmbir.py
@@ -208,7 +208,7 @@ class SVMBIRExtendedLoss(Loss):
     def __init__(
         self,
         *args,
-        prox_kwargs: Optional[dict] = None,
+        prox_kwargs: dict = {},
         positivity: bool = False,
         W: Optional[Diagonal] = None,
         **kwargs,
@@ -225,7 +225,7 @@ class SVMBIRExtendedLoss(Loss):
                 :meth:`svmbir.recon` prox routine. Defaults to
                 {"maxiter": 1000, "ctol": 0.001}.
             positivity: Enforce positivity in the prox operation. The
-                loss is infinity if any element of the input is negative.
+                loss is infinite if any element of the input is negative.
         """
         super().__init__(*args, **kwargs)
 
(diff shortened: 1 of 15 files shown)


scico/linop/radon_svmbir.py
```diff
@@ -190,20 +192,28 @@ class SVMBIRWeightedSquaredL2Loss(WeightedSquaredL2Loss):
     of :class:`scico.linop.Diagonal`. If :math:`W` is None, it is set to
     :class:`scico.linop.Identity`.
 
+    The extended loss differs from a typical weighted squared :math:`\ell_2` loss
+    is the following ways.
+
     When ``positivity=True``, the prox projects onto the non-negative
     quadrant, but the the loss, :math:`\alpha l(\mb{y}, A(\mb{x}))`,
```
bwohlberg, line 199: quadrant --> orthant

scico/linop/radon_svmbir.py
```diff
@@ -190,20 +192,28 @@ class SVMBIRWeightedSquaredL2Loss(WeightedSquaredL2Loss):
...
     is unaffected by this setting and still evaluates to finite values
     when :math:`\mb{x}` is not in the non-negative quadrant.
+
+    When the ``is_masked`` option of the associated ParallelBeamProjector is True,
+    the reconstruction is computed over a masked region of the image as described
```
bwohlberg, line 204: Consider adding sphinx `:class:` cross-referencing so that `ParallelBeamProjector` links to the corresponding docs page for that class.

scico/linop/radon_svmbir.py
```diff
@@ -190,20 +192,28 @@ class SVMBIRWeightedSquaredL2Loss(WeightedSquaredL2Loss):
...
+    in the ParallelBeamProjector class.
```
bwohlberg, line 205: As above, possibly with rephrasing, e.g. in class :class:ParallelBeamProjector (backticks omitted due to conflict with markup here)

scico/linop/radon_svmbir.py
```diff
@@ -212,13 +222,10 @@ def __init__(
...
-                :meth:`svmbir.recon` prox routine. Note that omitting
-                this dictionary will cause the default dictionary to be
-                used, however omitting entries in the passed dictionary
-                causes the defaults of the underlying :meth:`svmbir.recon`
-                prox routine to be used.
+                :meth:`svmbir.recon` prox routine. Defaults to
+                {"maxiter": 1000, "ctol": 0.001}.
             positivity: Enforce positivity in the prox operation. The
-                loss is not affected.
+                loss is infinity if any element of the input is negative.
```
bwohlberg, line 228: Perhaps "loss is infinite"?

scico/linop/radon_svmbir.py
```diff
@@ -230,15 +237,35 @@ def __init__(
         if prox_kwargs is None:
             prox_kwargs = dict()
```
bwohlberg, line 238: Egregious nitpick: I think ```prox_kwargs = {}``` would be a more standard way of writing this.

scico/linop/radon_svmbir.py
```diff
@@ -230,15 +237,35 @@ def __init__(
...
+
         svmbir_prox_args = dict()
-        if "maxiter" in prox_kwargs:
-            svmbir_prox_args["max_iterations"] = prox_kwargs["maxiter"]
-        if "ctol" in prox_kwargs:
-            svmbir_prox_args["stop_threshold"] = prox_kwargs["ctol"]
+        if "maxiter" in default_prox_args:
+            svmbir_prox_args["max_iterations"] = default_prox_args["maxiter"]
+        if "ctol" in default_prox_args:
+            svmbir_prox_args["stop_threshold"] = default_prox_args["ctol"]
```
bwohlberg, line 247: This approach makes sense for just two options to be passed to svmbir, but if that were ever to expand, I would suggest switching to a loop over a dict mapping our parameter names to svmbir ones.

scico/linop/radon_svmbir.py
```diff
@@ -230,15 +237,35 @@ def __init__(
...
         self.positivity = positivity
 
+        if W is None:
+            self.W = Identity(self.y.shape)
+        elif isinstance(W, Diagonal):
+            if snp.all(W.diagonal >= 0):
+                self.W = W
+            else:
+                raise Exception(f"The weights, W.diagonal, must be non-negative.")
+        else:
```
bwohlberg, line 259: Perhaps just "The weights, W, must be ..." since W.diagonal is internal and not part of the interface?

bwohlberg (approved):
Looks good to me, but it would be best to get at least one other review.

scico/linop/radon_svmbir.py
smajee: Thanks. I will keep that in mind as we change the svmbir interface.

scico/linop/radon_svmbir.py
smajee: Fixed.

scico/linop/radon_svmbir.py
smajee: Fixed.

scico/linop/radon_svmbir.py
smajee: Fixed.

scico/linop/radon_svmbir.py
smajee: Fixed.

scico/linop/radon_svmbir.py
smajee: Fixed.

scico/linop/radon_svmbir.py
smajee: Fixed.

scico/linop/radon_svmbir.py
```diff
@@ -190,20 +192,28 @@ class SVMBIRWeightedSquaredL2Loss(WeightedSquaredL2Loss):
     of :class:`scico.linop.Diagonal`. If :math:`W` is None, it is set to
     :class:`scico.linop.Identity`.
 
+    The extended loss differs from a typical weighted squared :math:`\ell_2` loss
+    is the following ways.
```
tbalke, line 196: in the following ways?

scico/linop/radon_svmbir.py
tbalke: the the

scico/linop/radon_svmbir.py
```diff
@@ -190,20 +192,28 @@ class SVMBIRWeightedSquaredL2Loss(WeightedSquaredL2Loss):
...
+
     When ``positivity=True``, the prox projects onto the non-negative
     quadrant, but the the loss, :math:`\alpha l(\mb{y}, A(\mb{x}))`,
     is unaffected by this setting and still evaluates to finite values
     when :math:`\mb{x}` is not in the non-negative quadrant.
+
+    When the ``is_masked`` option of the associated ParallelBeamProjector is True,
```
tbalke: `True`

scico/linop/radon_svmbir.py
tbalke: :class:`ParallelBeamProjector`

tbalke (approved):
Approve upon the few minor corrections

scico/linop/radon_svmbir.py
```diff
@@ -190,20 +192,28 @@ class SVMBIRWeightedSquaredL2Loss(WeightedSquaredL2Loss):
...
+    When the ``is_masked`` option of the associated :class:`.ParallelBeamProjector` is True,
+    the reconstruction is computed over a masked region of the image as described
+    in class :class:`.ParallelBeamProjector`.
     """
 
     def __init__(
         self,
         *args,
-        prox_kwargs: dict = {"maxiter": 1000, "ctol": 0.001},
+        prox_kwargs: dict = {},
```
Michael-T-McCann, line 211: initializing with a mutable type is risky, I suggest `=None` then `if prox_kwargs is None:`.

scico/linop/radon_svmbir.py
```diff
@@ -190,20 +192,28 @@ class SVMBIRWeightedSquaredL2Loss(WeightedSquaredL2Loss):
...
 
+    The extended loss differs from a typical weighted squared :math:`\ell_2` loss
+    is the following ways.
+
     When ``positivity=True``, the prox projects onto the non-negative
-    quadrant, but the the loss, :math:`\alpha l(\mb{y}, A(\mb{x}))`,
+    orthant, but the the loss, :math:`\alpha l(\mb{y}, A(\mb{x}))`,
     is unaffected by this setting and still evaluates to finite values
-    when :math:`\mb{x}` is not in the non-negative quadrant.
+    when :math:`\mb{x}` is not in the non-negative orthant.
```
Michael-T-McCann, line 201: "nonnegative". Generally don't hyphenate prefixes (unless they are words themselves, or if needed for clarity, or...)

scico/linop/radon_svmbir.py
```diff
@@ -212,13 +222,10 @@ def __init__(
...
-                :meth:`svmbir.recon` prox routine. Note that omitting
-                this dictionary will cause the default dictionary to be
-                used, however omitting entries in the passed dictionary
-                causes the defaults of the underlying :meth:`svmbir.recon`
-                prox routine to be used.
+                :meth:`svmbir.recon` prox routine. Defaults to
+                {"maxiter": 1000, "ctol": 0.001}.
             positivity: Enforce positivity in the prox operation. The
-                loss is not affected.
+                loss is infinite if any element of the input is negative.
```
Michael-T-McCann, line 228: contradicts earlier statement of "unaffected"?

tbalke:
Perhaps we also want to add an example for the SVMBIRExtendedLoss?

scico/linop/radon_svmbir.py
```diff
@@ -227,18 +234,35 @@ def __init__(
...
 
         self.positivity = positivity
 
+        if W is None:
+            self.W = Identity(self.y.shape)
+        elif isinstance(W, Diagonal):
+            if snp.all(W.diagonal >= 0):
+                self.W = W
+            else:
+                raise Exception(f"The weights, W, must be non-negative.")
```
Michael-T-McCann, line 255: This is checked by the `__init__` of `WeightedSquaredL2Loss`. Would it be possible to reduce duplication by making use of `super().__init__`?

Michael-T-McCann (approved):
Approved pending a couple minor comments.

scico/linop/radon_svmbir.py
smajee: How about "... in the following aspects"?

scico/linop/radon_svmbir.py
tbalke: Check out `admm.py`

```python
    def __init__(self, cg_kwargs: Optional[dict] = None, cg_function: str = "scico"):
        """Initialize a :class:`LinearSubproblemSolver` object.

        Args:
            cg_kwargs: Dictionary of arguments for CG solver. See
                documentation for :func:`scico.solver.cg` or
                :func:`jax.scipy.sparse.linalg.cg`,
                including how to specify a preconditioner.
                Default values are the same as those of
                :func:`scico.solver.cg`, except for
                ``"tol": 1e-4`` and ``"maxiter": 100``.
            cg_function: String indicating which CG implementation to
                use. One of "jax" or "scico"; default "scico".  If
                "scico", uses :func:`scico.solver.cg`. If "jax", uses
                :func:`jax.scipy.sparse.linalg.cg`. The "jax" option is
                slower on small-scale problems or problems involving
                external functions, but can be differentiated through.
                The "scico" option is faster on small-scale problems, but
                slower on large-scale problems where the forward
                operator is written entirely in jax.
        """

        default_cg_kwargs = {"tol": 1e-4, "maxiter": 100}
        if cg_kwargs:
            default_cg_kwargs.update(cg_kwargs)
        self.cg_kwargs = default_cg_kwargs
```

scico/linop/radon_svmbir.py
smajee: Changed it.

scico/linop/radon_svmbir.py
smajee: Fixed.

scico/linop/radon_svmbir.py
smajee: Fixed

scico/linop/radon_svmbir.py
smajee: Fixed

scico/linop/radon_svmbir.py
smajee: Fixed

scico/linop/radon_svmbir.py
smajee: Good catch! Fixed it now.

scico/linop/radon_svmbir.py
smajee: The `SVMBIRExtendedLoss` inherits from `Loss` instead of `WeightedSquaredL2Loss` since it is not a true WeightedSquaredL2Loss. So `super().__init__` won't work here since `Loss` doesn't have the weight.

One Way to deduce duplication would be to define a `WeightedLoss ` that inherits from loss. Both `WeightedSquaredL2Loss` and `SVMBIRExtendedLoss` can inherit from this.

scico/linop/radon_svmbir.py
smajee: Fixed it.

smajee:
> Perhaps we also want to add an example for the SVMBIRExtendedLoss?

Talked with @bwohlberg  about it. He seemed to think that it wasn't necessary since there would be a lot of code duplication between the `SVMBIRExtendedLoss` and `SVMBIRWeightedSquaredL2Loss` examples.

PR #194, opened 2022-01-27
Fix scico.random failure with latest jax version

bwohlberg:
Fix scico.random failure with latest jax version.


scico/random.py
```diff
@@ -172,7 +171,8 @@ def _wrap(fun):
 
 def _is_wrappable(fun):
     params = inspect.signature(getattr(jax.random, fun)).parameters
-    return list(params.keys())[0] == "key" and "shape" in params.keys()
+    prmkey = list(params.keys())
+    return prmkey and prmkey[0] == "key" and "shape" in params.keys()
```
tbalke, line 175: I'd prefer
```
return prmkey and (prmkey[0] == "key") and ("shape" in params.keys())
```
to make it easier to read. Your call

tbalke (approved):

scico/random.py
bwohlberg: I see your point, but my impression is that standard Python style is somewhat different from other languages in this respect, in that redundant parentheses are discouraged. I'll leave it as it is for this PR, but let's discuss.

Issue #195, opened 2022-01-28
Dependencies for examples


bwohlberg:
Some notes on installing example dependencies have been added to `docs/source/examples.rst`. Since this file is auto-generated by `examples/makeindex.py`, they are removed whenever the index is update. Obvious solutions are either:
1. move the notes to a different file, or
2. include them in the `examples/makeindex.py` script.

PR #209, opened 2022-02-03
Improve `ray.tune` usage example.

bwohlberg:
Improve `ray.tune` usage example. Also:
- Some minor improvements to `scico.ray.tune` wrapper.
- Extend tests for  `scico.ray.tune`.
- Add `ray.tune` usage example to examples index.


docs/source/examples.rst
```diff
@@ -6,22 +6,6 @@ Usage Examples
...
-The additional requirements should be installed via pip, with the exception of ``astra-toolbox``,
-which should be installed via conda:
-
-   ::
-
-      conda install -c astra-toolbox astra-toolbox
-      pip install -r examples/examples_requirements.txt # Installs other example requirements
-
-The dependencies can also be installed individually as required.
-
```
smajee, line 24: Did you remove the example related dependencies? I don't think these are there anywhere else.

scico/ray/tune.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright (C) 2021 by SCICO Developers
+# Copyright (C) 2021-2022 by SCICO Developers
```
smajee, line 2: For other files, should the end year be the current year or the year the file was last edited on?
In the long-term, should we manually update these or should we make a script (say in scico/misc) that updates these automatically.

smajee (approved):
Looks good other than the concern about removal of example dependencies.

docs/source/examples.rst
bwohlberg: This is going to be a continual pain until #195 is addressed. I'll fix manually for now.

scico/ray/tune.py
bwohlberg: My understanding is that the end year should correspond to the year of most recent edit. I think we can live with manual updates for now.

Issue #212, opened 2022-02-07
Mysterious and intermittent test failures


Michael-T-McCann:
We have had occasional test failures: https://github.com/lanl/scico/runs/5062268572?check_suite_focus=true, https://github.com/lanl/scico/runs/5078490742?check_suite_focus=true

These involve `scico/test/test_functional.py`, and specifically `scico/test/test_functional.py::TestNormProx::test_prox[float32-0.01-NuclearNorm] and `scico/test/linop/test_optics.py::TestPropagator::test_prop_adjoint[FraunhoferPropagator-1]`. Search the very long stack trace for "typeerror:" to see what I think is the actual error.

Michael-T-McCann:
Minimal code to reproduce. In the file `temp_svd.py`, put

```python
import jax
import jax.numpy as jnp

import scico.numpy as snp
import scico
from scico.random import randn

def test_prox():
    v, key = randn(shape=(32, 1), key=None, seed=3)

    svdU, svdS, svdV = snp.linalg.svd(v, full_matrices=False)
    svdS = snp.maximum(0, svdS - 1.0)
    x = svdU @ snp.diag(svdS) @ svdV
    sigma = snp.linalg.svd(x, compute_uv=False)
    snp.sum(sigma)

test_prox()
```

then try
```bash
while python temp_svd.py
do
:
done
```

Michael-T-McCann:
I have opened an issue with jax team on this: https://github.com/google/jax/issues/9483

Michael-T-McCann:
When/if this gets fixed on jax's side, let's revisit (probably roll back) #214.

Michael-T-McCann:
Revert the changes.

PR #214, opened 2022-02-08
Change svd call to avoid compute_uv=False

Michael-T-McCann:
Intended as a temporary workaround for the intermittent test failures discussed in #212 .


bwohlberg (approved):
Looks good, but let's leave #212 open until the jax bug is resolved as a reminder of the need to revert this change when possible.

PR #218, opened 2022-02-09
Address excessive memory usage in `functional.NuclearNorm.__call__`

bwohlberg:
A minor change to the work-around for google/jax#9483 that takes advantage of the low frequency of occurrence of the bug, and avoids potentially very high memory requirements for non-square matrices by use of the `full_matrices=False`  option.

### DIFF
--- a/scico/functional/_norm.py
+++ b/scico/functional/_norm.py
@@ -273,9 +273,13 @@ class NuclearNorm(Functional):
     has_prox = True
 
     def __call__(self, x: Union[JaxArray, BlockArray]) -> float:
-        # Set computute_uv=True to work around
-        # https://github.com/google/jax/issues/9483
-        _, s, _ = snp.linalg.svd(x, compute_uv=True)
+        # Original implementation of this function was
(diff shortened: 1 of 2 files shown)


scico/functional/_norm.py
```diff
@@ -273,9 +273,13 @@ class NuclearNorm(Functional):
...
-        # Set computute_uv=True to work around
-        # https://github.com/google/jax/issues/9483
-        _, s, _ = snp.linalg.svd(x, compute_uv=True)
+        # Original implementation of this function was
+        #   return snp.sum(snp.linalg.svd(x, compute_uv=False))
+        # The implementation here is a temporary work-around due
+        # to the bug reported at https://github.com/google/jax/issues/9483
+        s = snp.linalg.svd(x, full_matrices=False, compute_uv=False)
+        if isinstance(s, tuple):
+            _, s, _ = snp.linalg.svd(x, full_matrices=False, compute_uv=True)
```
Michael-T-McCann, line 282: `s = s[1]` should work.

bwohlberg: Good point!

Issue #215, opened 2022-02-08
Unchecked assumptions on attribute type


bwohlberg:
This line implicitly assumes that `A` is a `LinearOperator`
https://github.com/lanl/scico/blob/2b7a693234a06f27bea7fc7d52f3a1e7b41c08ea/scico/loss.py#L195
but according to the typing on the `A` parameter of the class `__init__`, 
https://github.com/lanl/scico/blob/2b7a693234a06f27bea7fc7d52f3a1e7b41c08ea/scico/loss.py#L124
`A` may also be a function, which does not have an `adj` method. 

The `prox` method needs some additional checking to ensure that `A` is a `LinearOperator`, raising an exception if it is not.

tbalke:
Good catch. I'll fix it

PR #220, opened 2022-02-10
add check for A not linop in prox

tbalke:
See #215

Note that check for `self.has_prox`, namely that `W` is `Diagonal`, was redundant and so I removed it.

### DIFF
--- a/scico/loss.py
+++ b/scico/loss.py
@@ -154,7 +154,7 @@ class WeightedSquaredL2Loss(Loss):
             prox_kwargs = dict
         self.prox_kwargs = prox_kwargs
 
-        if isinstance(self.A, linop.Diagonal) and isinstance(self.W, linop.Diagonal):
+        if isinstance(self.A, linop.LinearOperator):
             self.has_prox = True
 
     def __call__(self, x: Union[JaxArray, BlockArray]) -> float:
@@ -163,6 +163,12 @@ class WeightedSquaredL2Loss(Loss):
     def prox(
         self, v: Union[JaxArray, BlockArray], lam: float, **kwargs
     ) -> Union[JaxArray, BlockArray]:
+        if not isinstance(self.A, linop.LinearOperator):
+            raise NotImplementedError(
+                f"prox is not implemented for {type(self)} when `A` is {type(A)}; "
+                "must be LinearOperator"
+            )
+
         if isinstance(self.A, linop.Diagonal):
             c = 2.0 * self.scale * lam
             A = self.A.diagonal
--- a/scico/test/test_functional.py
+++ b/scico/test/test_functional.py
@@ -344,7 +344,7 @@ class TestLoss:
     def test_squared_l2(self):
         L = loss.SquaredL2Loss(y=self.y, A=self.Ao)
         assert L.has_eval == True
-        assert L.has_prox == False  # not diagonal
+        assert L.has_prox == True
 
         # test eval
         np.testing.assert_allclose(L(self.v), 0.5 * ((self.Ao @ self.v - self.y) ** 2).sum())
@@ -375,7 +375,7 @@ class TestLoss:
     def test_weighted_squared_l2(self):
         L = loss.WeightedSquaredL2Loss(y=self.y, A=self.Ao, W=self.W)
         assert L.has_eval == True
-        assert L.has_prox == False  # not diagonal
+        assert L.has_prox == True
 
         # test eval
         np.testing.assert_allclose(


tbalke:
In `test_functional.py` we get the following functions failing:
```python
    def test_squared_l2(self):
        L = loss.SquaredL2Loss(y=self.y, A=self.Ao)
        assert L.has_eval == True
        assert L.has_prox == False  # not diagonal

    def test_weighted_squared_l2(self):
        L = loss.WeightedSquaredL2Loss(y=self.y, A=self.Ao, W=self.W)
        assert L.has_eval == True
        assert L.has_prox == False  # not diagonal
```
with
```
        assert L.has_eval == True
>       assert L.has_prox == False  # not diagonal
E       AssertionError: assert True == False
E        +  where True = <class 'scico.loss.WeightedSquaredL2Loss'>\nhas_eval = True\nhas_prox = True\n        .has_prox
```

However, `self.Ao` is a `LinearOperator` so in this case we indeed have a prox. Why are we asserting that `has_prox` is `False`?

bwohlberg:
> Note that check for `self.has_prox`, namely that `W` is `Diagonal`, was redundant and so I removed it.

In what way was it redundant?

tbalke:
> > Note that check for `self.has_prox`, namely that `W` is `Diagonal`, was redundant and so I removed it.
> 
> In what way was it redundant?

We had
```python
if isinstance(self.A, linop.Diagonal) and isinstance(self.W, linop.Diagonal):
    self.has_prox = True
```

At this point `isinstance(self.W, linop.Diagonal) == True` always because we set it to a `Diagonal` in case it is `None`.

bwohlberg:
> At this point `isinstance(self.W, linop.Diagonal) == True` always because we set it to a Diagonal in case it is None.

But what if `W` is not `None`, but is set to something other than a `linop.Diagonal`?

tbalke:
> > At this point `isinstance(self.W, linop.Diagonal) == True` always because we set it to a Diagonal in case it is None.
> 
> But what if `W` is not `None`, but is set to something other than a `linop.Diagonal`?

Then `raise TypeError(f"W must be None or a linop.Diagonal, got {type(W)}")`

bwohlberg:
> However, `self.Ao` is a `LinearOperator` so in this case we indeed have a prox. Why are we asserting that `has_prox` is `False`?

Previously (some months ago), the prox was only defined if the linear operator was diagonal. That changed when you (I think) added a CG solution for non-diagonal case. I would have assumed that this test was a remnant from before, but that doesn't explain why it only started failing now.

bwohlberg:
> > But what if `W` is not `None`, but is set to something other than a `linop.Diagonal`?
> 
> Then `raise TypeError(f"W must be None or a linop.Diagonal, got {type(W)}")`
> 
> So it is either `Diagonal` or `None` in which case we set it to `Identity` which is a `Diagonal`

OK, that makes sense. The full context is not apparent in the diff!

tbalke:
> > However, `self.Ao` is a `LinearOperator` so in this case we indeed have a prox. Why are we asserting that `has_prox` is `False`?
> 
> Previously (some months ago), the prox was only defined if the linear operator was diagonal. That changed when you (I think) added a CG solution for non-diagonal case. I would have assumed that this test was a remnant from before, but that doesn't explain why it only started failing now.

Correct. I think that the test fails because this is not an appropriate test. I suggest changing the test to the negation. If the A is a linop then `has_prox` should be true.

bwohlberg:
> Correct. I think that the test fails because this is not an appropriate test. I suggest changing the test to the negation. If the A is a linop then `has_prox` should be true.

That seems reasonable, but I'm a bit concerned that we don't understand why the tests only started failing *now*.

tbalke:
> > Correct. I think that the test fails because this is not an appropriate test. I suggest changing the test to the negation. If the A is a linop then `has_prox` should be true.
> 
> That seems reasonable, but I'm a bit concerned that we don't understand why the tests only started failing _now_.

The previous test asserted that `has_prox == False` (incorrectly) and the loss.py code set the `has_prox` attribute to `False` (incorrectly) when `A` was a linop.

Fixing the loss.py code made fixing the test code necessary.

bwohlberg:
That makes sense w.r.t. tests, but it is a bit strange that we did not previously noticed that the new prox capability that was added was crippled by this oversight. Nevertheless, you may as well go ahead and make the proposed changes to the tests.

bwohlberg (approved):
Looks good, but let's ask @Michael-T-McCann to take a look too.

PR #221, opened 2022-02-10
Add new example script

bwohlberg:
Add new example script.

### DIFF
--- a/examples/scripts/superres_ppp_dncnn_admm.py
+++ b/examples/scripts/superres_ppp_dncnn_admm.py
@@ -1,0 +1,135 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# This file is part of the SCICO package. Details of the copyright
+# and user license can be found in the 'LICENSE.txt' file distributed
+# with the package.
+
+"""
+Image Superresolution (ADMM Plug-and-Play Priors w/ DnCNN)
+==========================================================
+
+This example demonstrates the use of the ADMM Plug and Play Priors
+(PPP) algorithm :cite:`venkatakrishnan-2013-plugandplay2` for solving
+a simple image superresolution problem.
+"""
+
+import jax
+
+import scico
+import scico.numpy as snp
+import scico.random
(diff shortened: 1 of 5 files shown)


examples/scripts/superres_ppp_dncnn_admm.py
```diff
@@ -0,0 +1,135 @@
...
+
+"""
+Create a test image by downsampling and adding Gaussian white noise.
+"""
+rate = 4  # downsampling rate
+σ = 2e-2  # noise standard deviation
+
+Afn = lambda x: downsample_image(x, rate=rate)
+s = Afn(img)
+inshape = img.shape
```
tbalke, line 53: May as well call then `input_shape` and `output_shape`?

bwohlberg: Fair point; will do.

examples/scripts/superres_ppp_dncnn_admm.py
```diff
@@ -0,0 +1,135 @@
...
+fig = plot.figure(figsize=(8, 6))
+ax0 = plot.plt.subplot2grid((1, rate + 1), (0, 0), colspan=rate)
+plot.imview(img, title="Reference", fig=fig, ax=ax0)
+ax1 = plot.plt.subplot2grid((1, rate + 1), (0, rate))
+plot.imview(sn, title="Downsampled", fig=fig, ax=ax1)
+fig.show()
+
+
+"""
+Show recovered full-resolution images.
```
tbalke, line 124: I would show the ground truth image as well but just my humble opinion.

tbalke (approved):

examples/scripts/superres_ppp_dncnn_admm.py
bwohlberg: That's the "reference" image in the previous figure.

tbalke: Right, although I mean in the same figure as the reconstructions.

PR #223, opened 2022-02-12
Fix some typing and docstring issues

bwohlberg:
Fix some typing and docstring issues.

### DIFF
--- a/scico/examples.py
+++ b/scico/examples.py
@@ -12,7 +12,7 @@ import glob
 import os
 import tempfile
 import zipfile
-from typing import Optional
+from typing import List, Optional
 
 import numpy as np
 
@@ -242,7 +242,7 @@ def tile_volume_slices(x: Array, sep_width: int = 10) -> Array:
     return out
 
 
-def create_cone(img_shape: Shape, center: Optional[list[float]] = None) -> Array:
+def create_cone(img_shape: Shape, center: Optional[List[float]] = None) -> Array:
     """Compute a 2D map of the distance from a center pixel.
 
     Args:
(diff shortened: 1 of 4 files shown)


scico/examples.py
```diff
@@ -242,18 +242,20 @@ def tile_volume_slices(x: JaxArray, sep_width: int = 10) -> JaxArray:
...
+            the center of the image.
 
     Returns:
-        An image containing a 2D map of the distances
+        An image containing a 2D map of the distances.
     """
 
-    if center == None:
+    if center is None:
         center = [img_dim // 2 for img_dim in img_shape]
```
bwohlberg, line 259: Is the intent here really to only support integer center values? If not, this should be `... img_dim / 2 ...`.

smajee: Yes, I originally thought of only integer values. But making it floats (/ instead of //) would be more general and should work.

scico/examples.py
```diff
@@ -267,17 +269,18 @@ def create_cone(img_shape: Shape, center: Optional[list] = None):
...
 
     Args:
-        img_shape : Shape of the phontom to be created
-        radius_list : List of radii of the rings in the phantom
-        val_list : List of intensity values of the rings in the phantom
-        center : Tuple of center pixel ids. If None, this is set to the center of the image
+        img_shape: Shape of the phantom to be created.
+        radius_list: List of radii of the rings in the phantom.
+        val_list: List of intensity values of the rings in the phantom.
+        center: Tuple of center pixel ids. If ``None``, this is set to
```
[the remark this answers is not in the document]
smajee: Yes, by ids I meant indices.

smajee (approved):
Looks good to me.

PR #229, opened 2022-02-23
Check for blockarray inputs in from_operator

Michael-T-McCann:
Check for blockarray inputs in `from_operator`.


bwohlberg (approved):
Approved modulo comment on error message.

PR #233, opened 2022-02-24
Improve documentation for `linop.optics`

bwohlberg:
Improve documentation for `linop.optics`.


scico/linop/optics.py
```diff
@@ -91,15 +111,19 @@ def __init__(
...
-            z: Propagation distance.
-            pad_factor: Amount of padding to apply in DFT step.
+               `len(input_shape)==2` the same sampling interval is applied
+               to both dimensions. If `dx` is a tuple, it must have same
+               length as `input_shape`, and corresponds to either
+               :math:`(\Delta x,)` or :math:`(\Delta x, \Delta y)`.
+            k0: Illumination wavenumber, :math:`k_0`, corresponding to
+               :math:`2 \pi` / wavelength.
+            z: Propagation distance, :math:`z`.
+            pad_factor: The padded input shape is the input shape
```
bwohlberg, line 125: @lukepfister: Can you suggest some guidance to the user on selection of the padding factor?

scico/linop/optics.py
```diff
@@ -351,60 +381,60 @@ def adequate_sampling(self):
...
+    coordinates :math:`(k_x, k_y) = (\frac{k_0}{z} x_D,
+    \frac{k_0}{z} y_D)`.
 
-        Ignoring multiplicative prefactors, the Fraunhofer propagated
-        field is the Fourier transform of the source field, evaluated at
-        coordinates :math:`(k_x, k_y) = (\frac{k_0}{z} x_D,
-        \frac{k_0}{z} y_D)`.
+    In general, the sampling intervals (and thus plane lengths)
+    differ between source and destination planes. In particular,
+    :cite:`voelz-2011-computational`
```
bwohlberg, line 417: @lukepfister: Do these equations correspond to Eq. 5.18 in `voelz-2011-computational`?

lukepfister: yes

scico/linop/optics.py
lukepfister: That's a complicated subject;  it's the same as the choice of padding for DFT-based linear convolution but has extra physical interpretation.  Would probably just leave a pointer to Chapter 5 of voelz.

scico/linop/optics.py
bwohlberg: Thanks. Added a note to the docs.

scico/linop/optics.py
bwohlberg: Better than nothing, I guess, but I was hoping to provide some guidance that didn't require the user to first do enough reading to be able to write their own implementation!

lukepfister: 🤷‍♂️ I think it is too complicated to get into the docs. Unlike a standard analysis of convolution in a signal processing class, the side lengths have physical meaning. Sometimes the size of your domain is just the size of your domain and any spatial aliasing is really indicative that you haven’t modeled the transmission channel properly (not accounting for reflections, etc) 

there is a case to be made that those options should be removed entirely and the user should set up the domain side lengths to be larger.

scico/linop/optics.py
```diff
@@ -144,7 +144,7 @@ def __init__(
         self.k0: float = k0
         #: Shape of input after padding
         self.padded_shape: Shape = tuple(pad_factor * s for s in input_shape)
-        #: Padded source plane side length `(dx[i] * padded_shape[i])`
+        #: Padded source plane side length (dx[i] * padded_shape[i])
```
Michael-T-McCann, line 160: Why the colons at the start of the comment?

Michael-T-McCann (approved):

scico/linop/optics.py
bwohlberg: Apparently it's sphinx syntax for [documenting attributes](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autoattribute).

PR #243, opened 2022-03-07
Make bm3d dependency optional

bwohlberg:
* Make bm3d dependency optional in the sense that only functions and scripts that make direct use thereof cannot be used when it isn't installed
* Some minor packaging improvements

### DIFF
--- a/scico/denoiser.py
+++ b/scico/denoiser.py
@@ -12,7 +12,12 @@ import numpy as np
 
 from jax.experimental import host_callback as hcb
 
-import bm3d as tunibm3d
+try:
+    import bm3d as tunibm3d
+except ImportError:
+    have_bm3d = False
+else:
+    have_bm3d = True
 
 import scico.numpy as snp
 from scico._flax import DnCNNNet, load_weights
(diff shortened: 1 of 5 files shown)


scico/denoiser.py
```diff
@@ -38,19 +43,21 @@ def bm3d(x: JaxArray, sigma: float, is_rgb: bool = False):
     Returns:
         Denoised output.
     """
+    if not have_bm3d:
+        raise RuntimeError("Package bm3d is required for use of this function." "")
```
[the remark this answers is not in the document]
bwohlberg: No. Probably a vestige of black intervention. Correction pushed.

scico/denoiser.py
```diff
@@ -38,19 +43,21 @@ def bm3d(x: JaxArray, sigma: float, is_rgb: bool = False):
...
 
     if np.iscomplexobj(x):
-        raise TypeError(f"BM3D requries real-valued inputs, got {x.dtype}")
+        raise TypeError(f"BM3D requires real-valued inputs, got {x.dtype}")
 
     # Support arrays with more than three axes when the additional axes are singletons.
     x_in_shape = x.shape
 
-    if x.ndim < 2:
+    if isinstance(x.ndim, tuple) or x.ndim < 2:
```
tbalke, line 60: What is the idea behind checking x.ndim is tuple?

tbalke: Unless the type annotations are incorrect. Do we allow BlockArrays here?

bwohlberg: We don't allow BlockArrays, but in the absence of the additional test for a tuple, this test fails to raise the appropriate exception if one is passed as the argument.

tbalke (approved):

Michael-T-McCann (approved):

PR #247, opened 2022-03-10
Add doctest of modules to workflow, fix modules so they pass

Michael-T-McCann:
Also gets the "unit tests / pytest" action to correctly install astra-toolbox and therefore not skip `test_radon_astra.py.`

Addresses https://github.com/scici/scico/issues/233


Michael-T-McCann:
Any ideas on why the readthedocs build is failing? I notice it is (1) throwing `No module named 'ipython_genutils'` and (2) running python3.7. But I can't see the action to debug further and have no idea how or why my changes would impact it.

bwohlberg:
> @bwohlberg Any ideas on why the readthedocs build is failing? I notice it is (1) throwing `No module named 'ipython_genutils'` and (2) running python3.7. But I can't see the action to debug further and have no idea how my changes would impact it.

It doesn't seem to be related to anything in this PR. Presumably a recent change in one of the docs dependencies, possibly `nbsphinx`, has broken the build. I'll look into it.

conftest.py
```diff
@@ -0,0 +1,20 @@
+"""
```
bwohlberg, line 1: Does this file have to be in the repo root directory?

docs/source/conf.py
```diff
@@ -162,7 +162,7 @@ def patched_parse(self):
 # built documents.
 #
 # The short X.Y version.
-with open(os.path.join("../../scico", "__init__.py")) as f:
+with open(os.path.join(confpath, "../../scico", "__init__.py")) as f:
     version = parse(next(filter(lambda line: line.startswith("__version__"), f))).body[0].value.s
```
bwohlberg, line 166: Why was this change necessary?

Also, although not introduced in this PR, one could argue that it's a bit pointless to separate the final two path components, to be joined by `os.path.join`, when one of those components (`../../scico`) had non-os-independent path separation. Perhaps worth fixing this now, making it `os.path.join(confpath, "..", "..", "scico", "__init__.py")`?

setup.py
```diff
@@ -75,6 +76,8 @@ def get_init_variable_value(var):
         lines = f.readlines()
     extras_require[extras_label] = [line.strip() for line in lines if line[0:2] != "-r"]
 
+# PEP517 workaround
+site.ENABLE_USER_SITE = True
```
bwohlberg, line 81: What does this do?

conftest.py
Michael-T-McCann: I know it works in `/scico/` or in `/scico/scico/`. In the later case, sphinx was trying to make docs pages for it and that caused errors. I doesn't work in `/scico/test/`.

It's ugly, but I couldn't find a way around it.

docs/source/conf.py
Michael-T-McCann: I was getting an error in either the doc tests or the doc build without it, but I don't recall which.

Separators fix.

setup.py
Michael-T-McCann: Without it, I was getting "error: can't create or remove files in install directory" on `pip install -e .`. I tracked that down to https://www.scivision.dev/python-pip-devel-user-install/, I'll add the link to the comment.

Why are we installing scico in dev mode for the tests? I don't know.

Why were we not getting this error before? I don't know.

bwohlberg:
> It doesn't seem to be related to anything in this PR. Presumably a recent change in one of the docs dependencies, possibly `nbsphinx`, has broken the build. I'll look into it.

Fixed in #248.

setup.py
bwohlberg: I don't think there's a good reason for the `-e` for the tests, but I guess this is still useful if it's now necessary for enabling user-editable installation.

conftest.py
bwohlberg: Strange, the `pytest` docs imply that it should be in the `tests` directory.

bwohlberg: But that's for normal tests, not doctests. I don't see a way around this either.

bwohlberg (approved):

Issue #231, opened 2022-02-23
Add a padding linop


Michael-T-McCann:
Padding and cropping are useful operations, especially when using circular convolutions (i.e., CircularConvolve). Add a `Pad` linop and a `Crop` linop, make sure that `Pad.T == Crop`. Could internally call `snp.pad`.

PR #249, opened 2022-03-11
Add a mechanism to create linear operators from functions

### LINKED ISSUE #245: Add a transpose linop
A linop providing a wrapper around `numpy.transpose` (i.e. a permutation of axes of an arbitrary-dimensional array, not just swapping axes of a two-dimensional array) would be a useful companion to the existing `linop.Slice` linop.

Michael-T-McCann:
Best explained with an example:

``` python
    >>> Sum = linear_operator_from_function(snp.sum)
    >>> H = Sum((2, 10), axis=1)
    >>> H @ snp.ones((2, 10))
    DeviceArray([10., 10.], dtype=float32)
```

And we can use that inside `linop/_linop.py` to wrap whatever we want, right now that's

``` python
Transpose = linear_operator_from_function(snp.transpose, "Transpose")
Sum = linear_operator_from_function(snp.sum, "Sum")
Pad = linear_operator_from_function(snp.pad, "Pad")
```

Closes #245 #231.


scico/linop/_linop.py
```diff
@@ -339,3 +339,38 @@ def __init__(
...
+    OpClass = type(classname, (LinearOperator,), {"__init__": __init__})
+    __class__ = OpClass  # needed for super() to work
+
+    OpClass.__doc__ = f"Linear operator built from {f.__module__}.{f.__name__}."
+
+    return OpClass
+
+
+Transpose = linear_operator_from_function(snp.transpose, "Transpose")
+Sum = linear_operator_from_function(snp.sum, "Sum")
```
Michael-T-McCann, line 375: Currently conflicts with the already-defined `Sum`.

scico/linop/__init__.py
```diff
@@ -11,7 +11,17 @@
 
 # isort: off
 from scico._generic_operators import LinearOperator
-from ._linop import Diagonal, Identity, power_iteration, operator_norm, Sum, Slice, valid_adjoint
+from ._linop import (
+    Diagonal,
```
bwohlberg, line 15: No need to include `linear_operator_from_function`?

scico/numpy/__init__.py
```diff
@@ -155,7 +155,10 @@ def vdot(a, b):
         _not_implemented_functions.append((name, func))
 
 _attach_wrapped_func(
-    _not_implemented_functions, _not_implemented, module_name=sys.modules[__name__]
+    _not_implemented_functions,
+    _not_implemented,
+    module_name=sys.modules[__name__],
+    fix_mod_name=False,
```
bwohlberg, line 161: Why the change here?

scico/test/linop/test_linop.py
```diff
@@ -519,49 +519,6 @@ def sumtestobj(request):
     yield SumTestObj(request.param)
 
 
-sum_axis = [
```
bwohlberg, line 522: Tests for old `Sum` linop removed, but not replaced with test for the new version?

Michael-T-McCann: I believe they are redundant with tests for `snp.sum`, which is tested over in test_numpy.py.

scico/numpy/__init__.py
Michael-T-McCann: I was messing around with how `__module__` was set for `scico.numpy` functions to get nicer docs. But in the end I didn't change it. The change here is just that I was explicit with the `fix_mod_name=False`, which is already the default.

scico/linop/__init__.py
Michael-T-McCann: Fixed.

scico/linop/_linop.py
```diff
@@ -339,3 +303,62 @@ def __init__(
 
     def _eval(self, x: JaxArray) -> JaxArray:
         return x[self.idx]
+
+
+def linear_operator_from_function(f: Callable, classname: str, f_name: Optional[str] = None):
```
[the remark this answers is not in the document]
Michael-T-McCann: Fixed.

scico/test/linop/test_linop.py
```diff
@@ -519,49 +519,6 @@ def sumtestobj(request):
     yield SumTestObj(request.param)
```
bwohlberg, line 519: Is this fixture still used by something?

scico/test/linop/test_linop.py
```diff
@@ -609,3 +566,30 @@ def test_slice_blockarray(idx):
     x = BlockArray.array((snp.zeros((3, 4)), snp.ones((3, 4, 5, 6))))
     A = linop.Slice(idx=idx, input_shape=x.shape, input_dtype=x.dtype)
     assert (A @ x).shape == x[idx].shape
+
+
+def test_transpose():
+    shape = (1, 2, 3, 4)
```
bwohlberg, line 572: Why these two new tests, but no corresponding new one for `Sum`?

scico/test/linop/test_linop.py
bwohlberg: But there are tests for the new Sum and Transpose linops. Why does a similar argument not apply to them?

scico/test/linop/test_linop.py
Michael-T-McCann: `snp.sum` is tested as part of the reductions tests in `test_numpy`. These two aren't (because they aren't reductions). I can see a case for leaving them out entirely, given that what is really being tested here is the underlying `jax.numpy` function and `jax.linear_transpose` (and all the plumbing that gets those methods called).

scico/test/linop/test_linop.py
Michael-T-McCann: Good catch, removed.

bwohlberg (approved):

bwohlberg:
Approved. I notice that `pytest-latest-jax` is failing, presumably due to some change in `jax` 0.3.2 released yesterday. Perhaps worth including a fix for that in this PR if it looks like a quick fix.

Issue #244, opened 2022-03-07
Docs for `scico.linop.svmbir` are inadequate


bwohlberg:
The docs for `scico.linop.svmbir` need improvement. Examples of important information that is currently omitted are the supported input shapes (2D or 3D) and the role of each axis in each case. Also, is ``num_channels`` really the "number of pixels in the sinogram"? Some of this documentation improvement can be achieved by reference to `svmbir` docs, but it should be via reference to specific functions rather than a link to the entire project.

smajee:
Issue #182 also has some comments stating similar docs inaccuracies. Would be good to cloese them together.

smajee:
Fixed in PR #252

Issue #241, opened 2022-03-03
Support fan beam geometry options of svmbir


bwohlberg:
There is now support for fan beam geometry in the svmbir pre-release branch (see cabouman/svmbir#231). The interface in `linop.radon_svmbir` should be extended to support it.

smajee:
As per discussion, ParallelBeamProjector should be renamed to TomographicProjector in order to support to fan beam.

smajee:
> This is urgent now since the recent `svmbir` release seems to have broken the`scico` interface to `svmbir`.

Got it.

smajee:
The urgent parts have been addressed via PR #246 https://scico.readthedocs.io/en/latest/examples/ct_svmbir_ppp_bm3d_admm_prox.html

smajee:
Fixed in PR #252

Issue #182, opened 2022-01-19
svmbir num_channels redundant?


smajee:
I noticed that the `num_channels` argument in scico-svmbir might be redundant. In the svmbir package the pixel-size and channel-size can be different and thus this parameter is needed. In scico, we simplify the interface to make them the same. Consequently, num_channels of the sinogram is always same as the corresponding dimension of the image. 

Any thoughts on this?

If we don't remove the parameter, I think we should at least correct the description which seems to be incorrect for the 3D case.

smajee:
As per our discussion we should expose svmbir's pixel size inputs in scico.

smajee:
Fixed in PR #252

PR #252, opened 2022-03-15
svmbir refactor

smajee:
- Generalize svmbir interface to include fan beam (solves issue #241)
- Add suitable defaults to fan-beam interface
- Add corresponding tests and an example
- Rebuild notebooks with new convention
- Improve svmbir docs (solves issue  #244 and #182)


examples/scripts/ct_fan_svmbir_ppp_bm3d_admm_prox.py
```diff
@@ -0,0 +1,252 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# This file is part of the SCICO package. Details of the copyright
+# and user license can be found in the 'LICENSE.txt' file distributed
+# with the package.
+
+"""
+CT Reconstruction (ADMM Plug-and-Play Priors w/ BM3D, SVMBIR+Prox)
```
bwohlberg, line 8: Title should be different from that of the corresponding parallel beam example.

examples/scripts/ct_fan_svmbir_ppp_bm3d_admm_prox.py
```diff
@@ -0,0 +1,252 @@
...
+==================================================================
+
+This example demonstrates the use of class
+[admm.ADMM](../_autosummary/scico.optimize.rst#scico.optimize.ADMM) to
+solve a tomographic reconstruction problem using the Plug-and-Play Priors
+framework :cite:`venkatakrishnan-2013-plugandplay2`, using BM3D
+:cite:`dabov-2008-image` as a denoiser and SVMBIR :cite:`svmbir-2020` for
+tomographic projection.
+
```
bwohlberg, line 17: This text should mention the use of the fan beam.

Is there really enough difference with the parallel beam version to justify a different example? (Is it just a change to `geometry="fan"` in a few places?)

examples/scripts/ct_fan_svmbir_ppp_bm3d_admm_prox.py
smajee: Fixed it.

examples/scripts/ct_fan_svmbir_ppp_bm3d_admm_prox.py
smajee: Fixed it. 
I agree there is not much difference. but it is more than changing the geometry flag to "fan".
There are new options `dist_source_detector` and `magnification`, change in view angle range, more padding in the ground truth, adding `delta_channel` and `delta_pixel` options in the direct svmbir call. (The last two will not be needed once svmbir implements the new defaults).

So I am not sure whether an example is needed or not.

scico/linop/radon_svmbir.py
```diff
@@ -30,39 +30,59 @@
...
     ):
         """
         Args:
-            input_shape: Shape of the input array.
+            input_shape: Shape of the input array. Can be of length
+                either 2 (2D array) or 3 (3D array). When specifying
+                a 2D array, the format for the input_shape is
+                (num_rows,num_cols). For a 3D array, the format for
+                the input_shape is (num_slices,num_rows,num_cols).
+                Note that (num_rows,num_cols) and (1,num_rows,num_cols)
```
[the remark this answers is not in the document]
bwohlberg: Both, I would say.

examples/scripts/ct_fan_svmbir_ppp_bm3d_admm_prox.py
smajee: So, with the same phantom, make projections and reconstruction with parallel beam and fan beam?

bwohlberg: Yes, that's what I was thinking. @Michael-T-McCann: Any thoughts?

smajee:
FYI, the code coverage was passing earlier. It is failing after expanding the docstrings

scico/linop/radon_svmbir.py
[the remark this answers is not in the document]
smajee: Added some explanations. After a careful look, I could not find places in the svmbir docs that explain the slices, rows, and cols properly

examples/scripts/ct_fan_svmbir_ppp_bm3d_admm_prox.py
smajee: I updated the example to compare fan-beam and parallel-beam reconstructions.

Michael-T-McCann: On the application side, you usually wouldn't get to choose whether to take fan beam or parallel beam measurements, so a comparison between the two may not be very meaningful. As an alternative, could you take measurements with a slight fan and compare fan and parallel reconstructions? This is something experimentalists do (reconstruct fan geometry with a parallel approximation).

Michael-T-McCann (approved):

examples/scripts/ct_fan_svmbir_ppp_bm3d_admm_prox.py
bwohlberg: Fair point, but I was primarily thinking of contrasting the actual usage of the class in the two different contexts rather than constructing a realistic example.

bwohlberg:
> FYI, the code coverage was passing earlier. It is failing after expanding the docstrings.

That happens sometimes. It's a really trivial decrease in coverage. I would ignore it.

examples/scripts/ct_fan_svmbir_ppp_bm3d_admm_prox.py
smajee: Made the change.

examples/scripts/ct_fan_svmbir_ppp_bm3d_admm_prox.py
```diff
@@ -0,0 +1,285 @@
...
+x_extloss_fan = solver_extloss_fan.solve()
+hist_extloss_fan = solver_extloss_fan.itstat_object.history(transpose=True)
+
+print()
+x_extloss_parallel = solver_extloss_parallel.solve()
+hist_extloss_parallel = solver_extloss_parallel.itstat_object.history(transpose=True)
+
+
+"""
+Show the recovered images.
```
bwohlberg, line 210: Worth adding a note here explaining why the parallel beam reconstructions is so bad. The reason is alluded to in the header comment, but a reminder here wouldn't hurt.

scico/linop/radon_svmbir.py
```diff
@@ -30,39 +30,65 @@
...
 
 
-class ParallelBeamProjector(LinearOperator):
-    r"""Parallel beam Radon transform based on svmbir.
+class TomographicProjector(LinearOperator):
+    r"""Tomographic projector based on svmbir.
 
     Perform tomographic projection of an image at specified angles, using
     the `svmbir <https://github.com/cabouman/svmbir>`_ package. The
     ``is_masked`` option selects whether a valid region for projections
```
bwohlberg, line 38: Wrong markup: should be single quotes for most of the places (other than None and True) where double quotes are currently used.

scico/linop/radon_svmbir.py
```diff
@@ -30,39 +30,65 @@
...
-            input_shape: Shape of the input array.
+            input_shape: Shape of the input array. Can be of length
+                either 2 (2D array) or 3 (3D array). When specifying
+                a 2D array, the format for the input_shape is
+                ``(num_rows,num_cols)``. For a 3D array, the format for
+                the input_shape is ``(num_slices,num_rows,num_cols)``.
+                Here ``num_slices`` denotes the number of slices in the
+                input, ``num_rows`` and ``num_cols`` denote the number
+                of rows and columns in a single slice of the input.
+                A slice is defined as the plane perpendicular to the
```
bwohlberg, line 69: Good to have slice defined here, but it would also help to explain how row and column fit into this geometry.

examples/scripts/ct_fan_svmbir_ppp_bm3d_admm_prox.py
smajee: Fixed it.

scico/linop/radon_svmbir.py
smajee: Fixed it.

scico/linop/radon_svmbir.py
```diff
@@ -87,6 +113,28 @@ def __init__(
         else:
             self.roi_radius = max(self.svmbir_input_shape[1], self.svmbir_input_shape[2])
 
+        self.geometry = geometry
+        self.dist_source_detector = dist_source_detector
+        self.magnification = magnification
+
+        if self.geometry == "fan":
+            if self.dist_source_detector is None:
+                raise AssertionError("dist_source_detector must be specified if geometry is fan")
```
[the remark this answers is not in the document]
smajee: Fixed it.

scico/linop/radon_svmbir.py
smajee: I have put a basic description, but I am unsure about a more detailed description. I am not sure whether the rows or columns are oriented along X-rays. The svmbir documentation also seems to be missing this piece of information.

Is there anything else I could add here to make it clearer?

examples/scripts/ct_fan_svmbir_ppp_bm3d_admm_prox.py
Michael-T-McCann: I can't readily see the reconstructions. I think this would be most convincing if you pick a small cone angle such that the parallel reconstruction is a bit distorted, but not total nonsense.

smajee: Is there some issue with the plot display? Works for me, but I run it in as a notebook (due to lack of admin access).

bwohlberg: I think @Michael-T-McCann's point is that the reconstruction quality is really bad, not that he can't see the plots at all.

smajee: That is a good point. I have reduced the cone angle so that the parallel beam approximation is better and thus the parallel-beam reconstruction also better.

scico/linop/radon_svmbir.py
bwohlberg: Why do we use `(num_slices,num_rows,num_cols)` while `svmbir` uses `(num_views, num_slices, num_channels)`? I think their terminology is more standard for CT, and the axis names do have clearly defined meanings w.r.t. the imaging geometry.

bwohlberg:
Perhaps worth noting pending `svmbir` changes in cabouman/svmbir#237.

scico/linop/radon_svmbir.py
smajee: svmbir also uses the terminology `(num_slices,num_rows,num_cols)` for the reconstruction. The `(num_views, num_slices, num_channels)` terminology is used for the sinograms.

bwohlberg: OK, I didn't notice that.

scico/linop/radon_svmbir.py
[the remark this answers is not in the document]
bwohlberg: It would be really helpful to the user to have some understanding of how the reconstruction shape relates to the sinogram shape. Perhaps ask the svmbir developers or do a simple test to figure out what the relationship is?

scico/linop/radon_svmbir.py
```diff
@@ -30,39 +30,76 @@
...
+                perpendicular to the axis of rotation of the tomographic
+                system. Note that `input_shape=(num_rows, num_cols)` and
+                `input_shape=(1, num_rows, num_cols)` result in the
+                same underlying projector.
             angles: Array of projection angles in radians, should be
                 increasing.
-            num_channels: Number of pixels in the sinogram.
+            num_channels: Number of detector channels in the sinogram
+                data.
             center_offset: Position of the detector center relative to
```
bwohlberg, line 89: It would help to explain where the center of rotation is.

scico/linop/radon_svmbir.py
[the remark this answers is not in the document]
smajee: Made the changes. Added the description of the output shape at the end of the docstring of the `__init__` function. Is there a better place to put it?

scico/linop/radon_svmbir.py
smajee: Added some extra explanation on the center of rotation. Did you mean something different?

bwohlberg: Looks good.

bwohlberg (approved):

Michael-T-McCann (approved):

PR #253, opened 2022-03-16
Replaced gpu<->cpu copies with host_callback in minimize.

### LINKED ISSUE #204: Replace manual GPU<->host copies in `scico.solver` with host callback
In order to use `scicpy.minimize`, we currently manual copy arrays back to the host 
https://github.com/lanl/scico/blob/ec3dea364b03b63578e962cbe90f0d25db7a4fb8/scico/solver.py#L164-L166
and send the result back to the GPU if needed
https://github.com/lanl/scico/blob/ec3dea364b03b63578e962cbe90f0d25db7a4fb8/scico/solver.py#L200-L201

FernandoDavis:
Added the host_callback functionality using a lambda expression for the minimizer function.  

- Return value of minimizer function is a non Jax-type object. To fix incompatibility with hcb, I only extracted the x value from the response object. 
- This meant a change in tests was needed from out.x to just out. 
- Did not find a suitable workaround to let me keep the response object while also allowing the hcb to work. Missing data from this new format is a success boolean to check if minimize worked, a message that explains if it did not, and some other attributes listed: fun, jax, hess, etc. 

Closes #204


FernandoDavis:
I'll check why the pytest is failing now in a few.

bwohlberg:
Ready for review?

FernandoDavis:
Yes! I'll be available if anything comes up.

scico/solver.py
```diff
@@ -229,25 +230,36 @@ def minimize(
...
         jac = False
 
-    res = spopt.minimize(
-        min_func,
-        x0=x0,
-        args=args,
-        jac=jac,
-        method=method,
-        options=options,
+    res_dict = {}
```
Michael-T-McCann, line 233: This might be cleaner using the following approach:
```python
    outside = None
    def f():
        nonlocal outside
        outside = 3
    print(outside)
    f()
    print(outside)
```

scico/solver.py
```diff
@@ -229,25 +230,36 @@ def minimize(
...
+
+    def fun(_):
```
Michael-T-McCann, line 235: Does `fun` need to have one argument for `hcb` to work? If so, `x0` could be a good choice. If not, remove `_`.

scico/solver.py
```diff
@@ -229,25 +230,36 @@ def minimize(
...
+        )  # Returns OptimizeResult
+        res_dict.update({"result": min_res})  # Updates res dict with side effect
+        return min_res.x.astype(x0.dtype)  # Return for host_callback
+
+    # hcb call with side effects to get the OptimizeResult on the same device it was called
+    hcb.call(
+        fun,
+        arg=None,
+        result_shape=x0,
+        call_with_device=isinstance(x0, jax.interpreters.xla.DeviceArray),
```
Michael-T-McCann, line 252: What is this doing? "`call_with_device` – if True then the callback function is invoked with the device from which the call originates as a keyword argument." But `fun` doesn't do anything with its arguments, right?

scico/solver.py
```diff
@@ -229,25 +230,36 @@ def minimize(
...
     )
 
+    res = res_dict["result"]
     # un-vectorize the output array, put on device
     res.x = snp.reshape(
         res.x, x0_shape
     )  # if x0 was originally a BlockArray then res.x is converted back to one here
 
     res.x = res.x.astype(x0_dtype)
```
Michael-T-McCann, line 261: Duplicates line 245

scico/solver.py
FernandoDavis: The call parameter arg requires something to be there so I made the adjustment to use x0, thanks!

scico/solver.py
FernandoDavis: This line is so that I could extract the side effect created by fun out of the dictionary for readability (using res instead of res_dict["result"] in the later parts of the method. Should I still remove it and fix it to use that?

scico/solver.py
FernandoDavis: Actually, I have to look into this again due to the change I did to the device input of fun. What I want to achieve with this is that when I call the hcb function, even though it doesn't actually do anything with the argument, the execution of the hcb returns the output to the same device it was called from and not to the default device.

FernandoDavis: I believe I fixed it, I will push what I have right now and test it from an environment with more than one device to ensure it's working properly. I will add some tests that reflect the device effect as well.

scico/solver.py
FernandoDavis: On it!

FernandoDavis: Worked just fine, thanks!

scico/solver.py
FernandoDavis: Addressed in the other comment, cleaned up syntax.

scico/solver.py
FernandoDavis: Added.

scico/solver.py
FernandoDavis: MyPy is complaining about res not having an attribute when overriden with the actual OptimizeResult

should I do: 

res = spopt.OptimizeResult({'x':""})

or should I revert to using the dictionary approach for the side effect?

Michael-T-McCann (approved):

FernandoDavis:
Yes! I'll be available if anything comes up.
________________________________

PR #255, opened 2022-03-18
Fix bugs in `loss` module

bwohlberg:
Fix some bugs/oversights in initialization of linear operators.

### DIFF
--- a/scico/loss.py
+++ b/scico/loss.py
@@ -403,12 +403,12 @@ def _dep_cubic_root(p, q):
     (see Sec. 3.C of :cite:`soulez-2016-proximity`).
     """
     q2 = q / 2
-    Δ = q2**2 + (p / 3) ** 3
+    Δ = q2 ** 2 + (p / 3) ** 3
     Δrt = snp.sqrt(Δ + 0j)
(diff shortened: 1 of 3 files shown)


Michael-T-McCann (changes requested):
These changes look fine, but I don't see any changes to the `linop` module, which is at odds with the name of the PR. Is something missing here?

bwohlberg:
> These changes look fine, but I don't see any changes to the `linop` module, which is at odds with the name of the PR. Is something missing here?

You're right. The change affected linop initialization, but was actually in the `loss` module. Title corrected accordingly.

scico/loss.py
```diff
@@ -58,7 +58,8 @@ def __init__(
         Args:
             y: Measurement.
             A: Forward operator. Defaults to ``None``, in which case
-               `self.A` is a :class:`.Identity`.
+               ``self.A`` is a :class:`.Identity` with input shape
```
FernandoDavis, line 61: Backticks in self.A should be reverted I believe

FernandoDavis (approved):
LGTM

Michael-T-McCann (approved):

PR #258, opened 2022-03-22
Add support for addition of functionals

bwohlberg:
Add support for addition of functionals. The result is a functional without a prox, but that can be evaluated if both summand functionals can be evaluated.

### DIFF
--- a/scico/functional/_functional.py
+++ b/scico/functional/_functional.py
@@ -45,9 +45,7 @@ has_prox = {self.has_prox}
     def __mul__(self, other):
         if snp.isscalar(other) or isinstance(other, jax.core.Tracer):
             return ScaledFunctional(self, other)
-        raise NotImplementedError(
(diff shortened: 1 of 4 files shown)


scico/functional/_functional.py
```diff
@@ -52,6 +52,13 @@ def __mul__(self, other):
     def __rmul__(self, other):
         return self.__mul__(other)
 
+    def __add__(self, other):
+        if isinstance(other, Functional):
+            return FunctionalSum(self, other)
+        raise NotImplementedError(
+            f"Operation __add__ not defined between {type(self)} and {type(other)}"
```
Michael-T-McCann, line 59: The recommended way to do this is to return the constant `NotImplemented` rather than raise an exception; see https://docs.python.org/3/library/constants.html#NotImplemented

bwohlberg: Fixed. (I wouldn't be surprised if there are some other instances of this error floating around in other modules.)

Issue #239, opened 2022-03-03
BlockArray reductions: tuples for `axis` argument


Michael-T-McCann:
In _real scientific imaging code_ one of our team members has wanted to sum each block of a block array along 2 of their 3 dimensions. Can we extend the blockarray reduction syntax to allow this? Currently it seems to require multiple calls to `sum`.

Issue #238, opened 2022-03-03
blockarray docs inconsistency


Michael-T-McCann:
In the blockarray docs, the examples in https://scico.readthedocs.io/en/latest/_autosummary/scico.blockarray.html#numpy-ufuncs don't seem to match between code and math, especially the `snp.sum(x, axis=2)`. We should check these.

Also, why does/should `snp.sum(x, axis=2)` work at all when `snp.sum(x[1].shape, 1)` gives an error? Similarly, why does `snp.sum(x, axis=99)` work?

bwohlberg:
@Michael-T-McCann: was this not resolved by #259?

Issue #237, opened 2022-03-03
Improve `repr` for BlockArray


Michael-T-McCann:
Printing a blockarray currently shows the flattened version, this is confusing. Let's make it print something more like what is shown in the docs, e.g.

```
BlockArray([[2, 2],
            [4,] ])

```

This involves the code at https://github.com/lanl/scico/blob/67da91873987c07acd886b6cd0a45365ccbb9383/scico/blockarray.py#L867-L868

Issue #159, opened 2022-01-06
Consistent broadcasting for BlockArray


Michael-T-McCann:
Arithmetic operations (+, -, *, ...) between a SCICO BlockArray and a NumPy ndarray are defined in the [BlockArray docs](https://scico.readthedocs.io/en/latest/_autosummary/scico.blockarray.html#module-scico.blockarray) to work in two cases: (1) when the ndarray is 1d and has size (i.e., number of elements) equal to the number of blocks; (2) when the ndarray has size equal to the size of the BlockArray.

This definition may be over restrictive, e.g.,
```
x = np.zeros(
    (3,)
)
y = BlockArray.zeros(
    ((1,),)
)
x + y  # error
```
    
And maybe isn't implemented correctly, e.g.,
```
x = np.zeros(
    (3, 4)
)
y = BlockArray.zeros(
    ((6,), (6,))
)
assert x.size == y.size
x + y  # error
```

Ideally, we could find a simple principle that would be intuitive and cover all relevant cases. One approach would be to generalize NumPy's scalar + ndarray behavior. That is, consider the ndarray to have been expanded (by duplication) into a BlockArray with the same number of blocks as the BlockArray, then perform the operation on these BlockArrays.

Examples:
```
np.zeros( (3,) ) + BlockArray.zeros( ((1,),) )  # output shape: ((3,),) + ((1,),) = ((3,),) 
```

Michael-T-McCann:
Confirmed, this has been fixed by the new version of BlockArray. The provided example now works:
```python
(snp.zeros( (3,) ) + snp.zeros( ((1,),) )).shape
((3,),)
```

PR #259, opened 2022-03-22
Simplify `BlockArray` implementation

### LINKED ISSUE #179: Consider simplification or replacement of `BlockArray`
`BlockArray`s are/have been at the center of several issues (#173 #159 #149 #147 #146), are difficult to maintain (because they touch jax internals), and necessitate a large amount of code bloat in SCICO (`scico.numpy`, `scico.scipy`).

Let's use this issue to discuss possible replacements `BlockArray`. What functionality would this replacement need to provide? How could we go about it (implementation-wise)?

Michael-T-McCann:
This is a work in progress PR to change the underlying implementation of `BlockArray` from a flattened `DeviceArray` to a list of `DeviceArray`s. The goal is to simplify the `BlockArray` implementation, reduce its coupling to jax internals, and provide additional functionality (mixed datatypes, blocks on different GPUs).

Closes #179. Also touches #237 #238 #159 #239.

#### Timing examples (best of 3 runs, total time, scripts edited to remove input, MacBook Pro, CPU):
`time python examples/scripts/denoise_tv_iso_pgm.py  > /dev/null 2>&1`
 old: 2.936, new 7.185, 
w/ additional `@jit.jit`, old: 2.936, new: 2.923

`time python examples/scripts/sparsecode_poisson_pgm.py  > /dev/null 2>&1`
old: 12.964, new 12.663

#### Timing simple ops
* array with 10,000 blocks, `%timeit -n 3 -r 3 x = snp.ones(10000*((2, 2),))`
     - new
`2.03 s ± 26.8 ms per loop (mean ± std. dev. of 3 runs, 3 loops each)`
    - old
    `62.7 ms ± 178 µs per loop (mean ± std. dev. of 3 runs, 3 loops each)`
* jitted creation
`f = jax.jit(lambda: snp.ones(10000*((2, 2),)))`
`f()  # trigger jit`
`%timeit -n 3 -r 3 z = f()`
    - new
`36.6 ms ± 345 µs per loop (mean ± std. dev. of 3 runs, 3 loops each)`
    - old
 `33.3 µs ± 30.2 µs per loop (mean ± std. dev. of 3 runs, 3 loops each)`
* smaller creation
`%timeit -n 3 -r 3 x = snp.ones(5*((512, 512),))`
    - new
    `2.1 ms ± 176 µs per loop (mean ± std. dev. of 3 runs, 3 loops each)`
    - old
    `1.38 ms ± 125 µs per loop (mean ± std. dev. of 3 runs, 3 loops each)`
* multiplying blocks
`x = snp.ones(512*((512, 512),))`
`%timeit -n 3 -r 3 y = x @ x`
    - new
`406 ms ± 60.5 ms per loop (mean ± std. dev. of 3 runs, 3 loops each)`
    - old
    `2.21 s ± 55.7 ms per loop (mean ± std. dev. of 3 runs, 3 loops each)`
* reduction
      `%timeit -n 3 -r 3 z = snp.linalg.norm(x)`
     - new
     `567 ms ± 50.5 ms per loop (mean ± std. dev. of 3 runs, 3 loops each)`
     - old
`152 ms ± 9.02 ms per loop (mean ± std. dev. of 3 runs, 3 loops each)`
* jited reduction
`f = jax.jit(lambda x: snp.linalg.norm(x))`
`f(x)  # trigger jit`
`%timeit -n 3 -r 3 z = f(x)`
    - new
    `269 ms ± 1.6 ms per loop (mean ± std. dev. of 3 runs, 3 loops each)`
    - old
`147 ms ± 1.35 ms per loop (mean ± std. dev. of 3 runs, 3 loops each)`
* smaller reduction
`x = snp.ones(5*((512, 512),))`
`%timeit -n 3 -r 3 z = snp.linalg.norm(x)`
    - new
    `3.86 ms ± 266 µs per loop (mean ± std. dev. of 3 runs, 3 loops each)`
    - new (jit as above)
    `1.63 ms ± 135 µs per loop (mean ± std. dev. of 3 runs, 3 loops each)`
    - old
`1.49 ms ± 203 µs per loop (mean ± std. dev. of 3 runs, 3 loops each)`
    - old (jit as above)
 `1.65 ms ± 228 µs per loop (mean ± std. dev. of 3 runs, 3 loops each)`


scico/linop/_stack.py
```diff
@@ -85,7 +85,7 @@ def __init__(
     def _eval(self, x: JaxArray) -> Union[JaxArray, BlockArray]:
         if self.collapsable and self.collapse:
             return snp.stack([op @ x for op in self.ops])
-        return BlockArray.array([op @ x for op in self.ops])
+        return BlockArray([op @ x for op in self.ops])
```
Michael-T-McCann, line 88: Any opinions on the syntax `snp.BlockArray(stuff)` vs `snp.blockarray(stuff)`? The second one is little more like NumPy.

scico/__init__.py
```diff
@@ -48,6 +48,8 @@
     "custom_vjp",
 ]
 
+from . import random, linop
```
bwohlberg, line 51: This is unavoidable?

scico/array.py
```diff
@@ -58,37 +57,22 @@ def ensure_on_device(
...
             )
 
+        arrays[i] = jax.device_put(arrays[i])
+
     if len(arrays) == 1:
         return arrays[0]
     return arrays
 
 
-def no_nan_divide(
```
bwohlberg, line 76: Where did this go?

scico/linop/_stack.py
bwohlberg: I'd lean towards copying NumPy style, but this is worth a discussion.

scico/loss.py
```diff
@@ -202,7 +202,7 @@ def __init__(
             self.has_prox = True
 
     def __call__(self, x: Union[JaxArray, BlockArray]) -> float:
-        return self.scale * (self.W.diagonal * snp.abs(self.y - self.A(x)) ** 2).sum()
```
bwohlberg, line 205: This syntax is no longer possible? I guess that's OK as long as we clearly document what operations should be avoided to ensure array/blockarray compatibility.

scico/numpy/__init__.py
```diff
@@ -5,163 +5,45 @@
...
+""":class:`scico.numpy.BlockArray`-compatible
 versions of :mod:`jax.numpy` functions.
 
 This modules consists of functions from :mod:`jax.numpy` wrapped to
-support compatibility with :class:`scico.blockarray.BlockArray`. This
+support compatibility with :class:`scico.numpy.BlockArray`. This
 module is a work in progress and therefore not all functions are
 wrapped. Functions that have not been wrapped yet have WARNING text in
 their documentation, below.
 """
```
bwohlberg, line 16: But also some miscellaneous other stuff that was copied in from `scico.array`. Is this module really the best home for this stuff?

scico/__init__.py
Michael-T-McCann: Not needed. This should either be removed or expanded to cover other modules. It slightly bothers me that one can't, e.g., 
```python
import scico
x, _ = scico.random(...)
```

scico/array.py
Michael-T-McCann: `scico.numpy.util`. Not all of these guys needed to move, but those that use `BlockArray` seemed to need to.

scico/loss.py
Michael-T-McCann: We could make this work, but it complicates the API. Currently all `BlockArray` methods map over blocks, including this one. You only get special reduction behavior when using `snp` methods.

scico/numpy/__init__.py
Michael-T-McCann: Probably not. I can put some more thought into what should go where.

scico/test/test_new_blockarray.py
```diff
@@ -0,0 +1,78 @@
+import operator as op
```
bwohlberg, line 1: This is a temporary filename?

scico/__init__.py
bwohlberg: Worth discussing.

scico/loss.py
bwohlberg: Probably not worth it then.

scico/array.py
```diff
@@ -11,22 +11,21 @@
 from __future__ import annotations
```
bwohlberg, line 11: Perhaps move all remaining content from this module to `scico.numpy`?

scico/numpy/blockarray.py
```diff
@@ -0,0 +1,352 @@
...
+    k
+    for k, v in dict(inspect.getmembers(DeviceArray)).items()
+    if isinstance(v, Callable)
+    and k[0] != "_"
+    and k not in dir(BlockArray)
+    and k not in skip_methods
+]
+
+for method in da_methods:
+    setattr(BlockArray, method, _da_method_wrapper(method))
```
bwohlberg, line 352: Impressive reduction in length and complexity.

scico/numpy/blockarray.py
```diff
@@ -0,0 +1,352 @@
...
+da_props = [
+    k
+    for k, v in dict(inspect.getmembers(DeviceArray)).items()
+    if isinstance(v, property) and k[0] != "_" and k not in dir(BlockArray) and k not in skip_props
+]
+
+for prop_name in da_props:
+    setattr(BlockArray, prop_name, _da_prop_wrapper(prop_name))
+
+""" Wrap DeviceArray methods. """
```
bwohlberg, line 327: These section dividers don't work so well because of the single space above and double space below. I suspect that `black` is insisting on the double space below? Can a triple space be added above, or would that also be removed by `black`?

scico/test/test_numpy.py
```diff
@@ -48,6 +48,7 @@ def test_reshape_array():
     np.testing.assert_allclose(b.ravel(), c.ravel())
 
 
+@pytest.mark.skip  # do we care to support svd of matrix operator?
```
bwohlberg, line 51: I assume the intent is to discuss these questions and permanently remove the tests in case of affirmative decision?

scico/linop/_stack.py
Michael-T-McCann: I made both do the same thing, but tended to write `snp.blockarray(...)` wherever I could. By comparison, `np.ndarray` and `np.array` do different things, with `np.ndarray` documented as a low-level method.

scico/__init__.py
Michael-T-McCann: Removed.

scico/numpy/__init__.py
Michael-T-McCann: As discussed, all of this moved into `snp.util` and `snp` has a new docstring.

scico/test/test_new_blockarray.py
Michael-T-McCann: Fixed.

scico/numpy/blockarray.py
Michael-T-McCann: Changed to normal line comments, which seem to work better.

scico/linop/_stack.py
bwohlberg: I think that's OK, but perhaps add a note in the docs indicating that the `snp.blockarray` form is preferred.

bwohlberg (approved):

scico/numpy/util.py
```diff
@@ -192,15 +171,43 @@ def indexed_shape(shape: Shape, idx: ArrayIndex) -> Tuple[int, ...]:
...
 def is_nested(x: Any) -> bool:
     """Check if input is a list/tuple containing at least one list/tuple.
 
     Args:
         x: Object to be tested.
 
     Returns:
-        ``True`` if `x` is a list/tuple of list/tuples, otherwise
-        ``False``.
+        ``True`` if `x` is a list/tuple of list/tuples, ``False`` otherwise.
```
tbalke, line 210: do you mean `list/tuple of lists/tuples`?

scico/test/linop/test_diff.py
```diff
@@ -3,62 +3,39 @@
...
+def test_adjoint(input_shape, input_dtype, axes, jit):
     ndim = len(input_shape)
-    x, _ = randn(input_shape, dtype=input_dtype)
-
     if axes in [1, (1,)] and ndim == 1:
-        with pytest.raises(ValueError):
-            A = FiniteDifference(
-                input_shape=input_shape, input_dtype=input_dtype, axes=axes, append=append
-            )
+        pass
```
tbalke, line 35: Just checkin, was this pass-cause left here with intention? And if so, why not use `if not`?

tbalke (approved):

scico/test/linop/test_diff.py
Michael-T-McCann: Nice catch. No intention, probably just quickly hacking the old test. Should be more readable now.

PR #260, opened 2022-03-23
Sphinx autodoc configuration changes

bwohlberg:
Disable inclusion of inherited members in sphinx autodoc. Also address some sphinx warnings and style issues.


Michael-T-McCann (approved):
Nice, looks like it really cleans up https://scico--260.org.readthedocs.build/en/260/_autosummary/scico.linop.html

crstngc (approved):
I agree. This is a nice change that reduces the clutter in the documentation.

PR #262, opened 2022-03-23
Update and fix conda environment construction script

bwohlberg:
* Remove options for installation with GPU support
* Remove requirement that script be run from repo root directory
* Updates to restore functionality broken by recent changes in the repo

### DIFF
--- a/misc/conda/make_conda_env.sh
+++ b/misc/conda/make_conda_env.sh
@@ -52,7 +52,7 @@ EOF
 
 
 OPTIND=1
-while getopts ":hyp:e:j:" opt; do
+while getopts ":hyp:e:" opt; do
     case $opt in
 	p|e) if [ -z "$OPTARG" ] || [ "${OPTARG:0:1}" = "-" ] ; then
 		     echo "Error: option -$opt requires an argument" >&2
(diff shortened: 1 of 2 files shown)


misc/conda/make_conda_env.sh
```diff
@@ -251,9 +219,16 @@ if [ "$(which dpkg 2>/dev/null)" ]; then
     fi
 fi
 
+echo
 echo "Activate the conda environment with the command"
 echo "  conda activate $ENVNM"
 echo "The environment can be deactivated with the command"
 echo "  conda deactivate"
+echo
+echo "Jax installed without GPU support. To avoid warning messages,"
```
tbalke, line 228: This means if GPU support is required then you should not use this script?

bwohlberg: No, it means that manual action is required *after* use of the script. I had assumed that this was implied by the following sentence "To include GPU support ...", but perhaps it's not as clear as I had thought?

tbalke: It was not clear to me. Let's make it more clear.

Also in the `.rst` file we could make it something like `To include GPU support, follow the jax installation instructions after running the installation script.`

bwohlberg: Done.

tbalke (approved):

Michael-T-McCann (approved):
Always like to see the negative lines of code added.

PR #273, opened 2022-04-08
Make compatible with jax v0.3.5

Michael-T-McCann:
The main problem was that `x.copy()` no longer turns a `DeviceArray` into an `ndarray`. Instead, use `np.array(x)` or `x.from_device()`.

### DIFF
--- a/scico/loss.py
+++ b/scico/loss.py
@@ -405,12 +405,12 @@ def _dep_cubic_root(p, q):
     (see Sec. 3.C of :cite:`soulez-2016-proximity`).
     """
     q2 = q / 2
-    Δ = q2 ** 2 + (p / 3) ** 3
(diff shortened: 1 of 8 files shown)


scico/loss.py
```diff
@@ -13,6 +13,8 @@
 from functools import wraps
 from typing import Callable, Optional, Union
 
+import jax
```
bwohlberg, line 16: No other changes to this module, but now `jax` is a required import? That's rather strange.

bwohlberg (approved):

scico/loss.py
Michael-T-McCann: Yup, it's a code smell that it was never a problem before. The important line:

```python
if snp.isscalar(other) or isinstance(other, jax.core.Tracer):
```

PR #275, opened 2022-04-10
Reduce test run time

bwohlberg:
Attempt to reduce test run time.


.github/workflows/pytest.yml
```diff
@@ -2,47 +2,98 @@
...
-
+  build:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-latest
+            label: linux-64
+            prefix: /usr/share/miniconda3/envs/test-env
+          - os: macos-latest
```
Michael-T-McCann, line 25: Does this double the time tests take? I would consider mac compatibility issues to be relatively unimportant, maybe worth a weekly/monthly test.

Michael-T-McCann (approved):

PR #278, opened 2022-04-15
Improve `SquaredL2SquaredAbsLoss`

bwohlberg:
Improve algorithm, docs, and testing of the cubic equation root calculation used by `SquaredL2SquaredAbsLoss`.

### DIFF
--- a/scico/test/functional/test_loss.py
+++ b/scico/test/functional/test_loss.py
@@ -212,8 +212,8 @@ class TestAbsLoss:
 
 def test_cubic_root():
     N = 10000
-    p, _ = uniform(shape=(N,), dtype=snp.float32, minval=-10.0, maxval=10.0)
-    q, _ = uniform(shape=(N,), dtype=snp.float32, minval=-10.0, maxval=10.0)
+    p, key = uniform(shape=(N,), dtype=snp.float32, minval=-10.0, maxval=10.0, seed=1234)
+    q, _ = uniform(shape=(N,), dtype=snp.float32, minval=-10.0, maxval=10.0, key=key)
     r = loss._dep_cubic_root(p, q)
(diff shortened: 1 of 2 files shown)


scico/test/functional/test_loss.py
```diff
@@ -208,3 +208,12 @@ def test_prox(self, loss_tuple):
 
         pf = prox_test((1 + 1j) * snp.zeros(self.v.shape), L, L.prox, 0.0)  # complex zero v
         pf = prox_test((1 + 1j) * snp.zeros(self.v.shape), L, L.prox, 1.0)  # complex zero v
+
+
+def test_cubic_root():
+    N = 10000
+    p, _ = uniform(shape=(N,), dtype=snp.float32, minval=-10.0, maxval=10.0)
```
Michael-T-McCann, line 215: consider seeding rng here so the test result is not stochastic

scico/test/functional/test_loss.py
```diff
@@ -208,3 +208,12 @@ def test_prox(self, loss_tuple):
...
+    q, _ = uniform(shape=(N,), dtype=snp.float32, minval=-10.0, maxval=10.0)
```
Michael-T-McCann, line 216: jax gotcha: I think p and q will be equal if you make them this way. Need to get the key from the first call and pass it to the second.

Michael-T-McCann (changes requested):

scico/test/functional/test_loss.py
bwohlberg: Fixed.

scico/test/functional/test_loss.py
bwohlberg: Indeed they were. Fixed.

Michael-T-McCann (approved):

PR #285, opened 2022-05-02
Restore missing `BlockArray` docs

bwohlberg:
A side of effect of #259 was removal of the detailed `BlockArray` docs (see [readthedocs](https://scico.readthedocs.io/en/latest/_autosummary/scico.numpy.blockarray.html)). Fixing this required
1. Moving `scico.numpy.blockarray` to `scico.numpy._blockarray` since the `blockarray = BlockArray.blockarray` line in `scico/numpy/__init__.py` overwrote the use of `blockarray` as a submodule name.
2. Moving the detailed docs into a Sphinx `.rst` file since there were additional difficulties in including the docs from `scico.numpy._blockarray`.

Some attention is still required, as indicated in comments to be added after creation of the PR.


docs/source/blockarray.rst
```diff
@@ -0,0 +1,198 @@
...
+     >>> x1, _ = scico.random.randn((16,), key=key)
+     >>> X = snp.blockarray((x0, x1))
+     >>> X.shape
+     ((32, 32), (16,))
+     >>> X.size
+     (1024, 16)
+     >>> len(X)
+     2
+
+While :func:`.snp.blockarray` will accept either :class:`~numpy.ndarray` or
```
bwohlberg, line 156: @Michael-T-McCann: This sentence (specifically the "backed by" phrasing) seems like a hold-over from the previous `BlockArray` implementation.

scico/numpy/_blockarray.py
```diff
@@ -0,0 +1,176 @@
...
+
+from jaxlib.xla_extension import DeviceArray
+
+from ._wrapped_function_lists import binary_ops, unary_ops
+
+
+class BlockArray(list):
+    """Block array class.
+
+    See :ref:`detailed BlockArray documentation <blockarray_class>`.
```
bwohlberg, line 25: @Michael-T-McCann: This is a bit ugly. Perhaps worth adding a brief summary here, in addition to the cross-reference to the full docs?

docs/source/blockarray.rst
```diff
@@ -0,0 +1,198 @@
...
+    >>> x = BlockArray((
+    ...     [[1, 3, 7],
+    ...      [2, 2, 1]],
+    ...     [2, 4, 8]
+    ... ))
+
+    >>> x.shape  # returns tuple
+    ((2, 3), (3,))
+
+    >>> x * 2  # returns BlockArray
```
bwohlberg, line 34: This doctest fails for reasons that are not immediately obvious. Running the test within `ipython` gives the type expected as per stated test result.

docs/source/blockarray.rst
Michael-T-McCann: Updated. I'm not sure about whether a copy is involved in construction, so I removed that text. I would suspect there is not a copy.

scico/numpy/_blockarray.py
Michael-T-McCann: Added.

docs/source/blockarray.rst
Michael-T-McCann: The tests are run with `JAX_ENABLE_X64=True`, which changes the default dtype of arrays. We can work around it for these docs and I'll make an issue to actually fix it.

Michael-T-McCann (approved):
Thanks for catching this and fixing it!

docs/source/blockarray.rst
bwohlberg: By work around, do you mean ignore the error and merge, or hold the merge for a fix in a separate PR?

PR #290, opened 2022-05-05
Various improvements

bwohlberg:
Split `admm` and `pgm` modules into multiple source files for easier editing, and various other minor improvements.


scico/optimize/_admm.py
```diff
@@ -0,0 +1,390 @@
...
+            subproblem_solver = GenericSubproblemSolver()
+        self.subproblem_solver: SubproblemSolver = subproblem_solver
+        self.subproblem_solver.internal_init(self)
+
+        # iteration number and time fields
+        itstat_fields = {
+            "Iter": "%d",
+            "Time": "%8.2e",
+        }
+        itstat_attrib = ["itnum", "timer.elapsed()"]
```
Michael-T-McCann, line 156: This itstat initialization is always a stumbling block for me when reading the class. Could it be easily refactored into its own function/class?

Michael-T-McCann (approved):

scico/optimize/_admm.py
bwohlberg: Perhaps. It would probably have to be some sort of private auxiliary class rather than a general utility, but that's worth considering. I'll make a separate issue as a reminder.

PR #292, opened 2022-05-09
Added docker file and instructions for docker img.

FernandoDavis:
- Dockerfile is used to create the base image and upload it to DockerHub.
- Installation instructions pull the image from DockerHub into a local dir
- Additionally, it runs the jupyter instance with the notebooks


bwohlberg:
I suggest moving `docs/Dockerfile` to `misc/docker/Dockerfile`.

bwohlberg:
If the CodeFactor complaints are meaningful, it would be worth trying to address them.

docs/Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM jupyter/base-notebook
+
+# Install git
+RUN conda update -n base conda
```
bwohlberg, line 4: Perhaps worth using `mamba` instead since it is supported by the base image?

bwohlberg:
Is it possible to construct an alternative docker image with GPU support, based either on one of the `gpu-jupyter` images, or perhaps directly on `nvidia/cuda`?

bwohlberg:
Is it possible to start `jupyter notebook` attached to `scico/data/notebooks/index.ipynb` rather than just in a listing of the directory `scico/data/notebooks`?

PR #293, opened 2022-05-09
Added the name of the BlockArray class to __repr__

FernandoDavis:
Added the name of the BlockArray class to __repr__


scico/numpy/_blockarray.py
```diff
@@ -79,6 +79,9 @@ def blockarray(iterable):
         """Construct a :class:`.BlockArray` from a list or tuple of existing array-like."""
         return BlockArray(iterable)
 
+    def __repr__(self):
+        return f"snp.BlockArray({super().__repr__()})"
```
Michael-T-McCann, line 83: @bwohlberg How about just `BlockArray(...` rather than `snp.BlockArray(...`? The former would match `DeviceArray`.

bwohlberg: Agreed.

Michael-T-McCann:
@bwohlberg Any ideas on these check failures?

bwohlberg:
I'm hoping the errors are transitory.

bwohlberg:
Looks good now.

bwohlberg:
@Michael-T-McCann: The current output of the code in this branch is not quite as outlined in #237. Did you envisage stripping the `DeviceArray` as in the example there, or leaving it in place as in the output for this branch?

Michael-T-McCann:
> Did you envisage stripping the `DeviceArray`

I did, but seeing how it looks I'm no longer sure. `BlockArray`s should only ever contain `DeviceArray`s so having "DeviceArray" in the repr is redundant. However, we do hope to support heterogeneous data types and having those types displayed in the repr would be useful. Removing "DeviceArray" but leaving the dtype looks odd to me:

```
    BlockArray([([[ 2,  6, 14],
		 [ 4,  4,  2]], dtype=...), ([ 4,  8, 16], dtype=...)])
```

I'm inclined to merge this as is.

PR #295, opened 2022-05-10
Improve example checking script

bwohlberg:
Improve example checking script and add github action to run it on PRs.


Michael-T-McCann (approved):
This works on my machine, which is a big improvement over the last version.

I did get `./scriptcheck.sh: line 113: kill: (95112) - No such process` after all the examples ran, not sure the cause.

PR #296, opened 2022-05-12
Modify `FiniteDifference` linop

bwohlberg:
Modify the `FiniteDifference` linop, changing the interface and adding support for more useful boundary handling options.


Michael-T-McCann (approved):
Looks like a nice solution.

PR #299, opened 2022-05-16
Improve handling of `h_center` parameter in `linop.CircularConvolve`

bwohlberg:
Improve handling of `h_center` parameter in `linop.CircularConvolve`. Allow tuples and lists instead of just DeviceArray type, as well as float or int for one-dimensional arrays. Also add tests for this parameter.


scico/linop/_circconv.py
```diff
@@ -124,7 +127,18 @@ def __init__(
             output_dtype = result_type(h.dtype, input_dtype)
 
             if self.h_center is not None:
-                offset = -self.h_center
+                if isinstance(self.h_center, DeviceArray):
+                    offset = -self.h_center
+                else:
+                    # support float or int values for h_center
+                    if isinstance(self.h_center, (float, int)):
+                        offset = -snp.array(
```
Michael-T-McCann, line 135: formatting looks slightly odd here but I assume that's black's doing

Michael-T-McCann (approved):

scico/linop/_circconv.py
bwohlberg: Do you mean the relatively deeply nested `if`/`else` structure? That's just a natural consequence of the actual logic here.

PR #301, opened 2022-05-20
Add Python 3.7 compatibility

Michael-T-McCann:
Three main changes:
- add `math.prod` if it doesn't exist
- add `typing.Literal` if it doesn't exist
- change syntax of `parse` call in `_version.py` if it fails

### DIFF
--- a/scico/__init__.py
+++ b/scico/__init__.py
@@ -12,6 +12,27 @@ __version__ = "0.0.3.dev0"
 
 import sys
 
+# Python 3.7 compatibility
+try:
+    from math import prod
+    from typing import Literal
+
+    del Literal, prod
+except ImportError:
+    import typing
+
+    typing.Literal = typing.Any
+
+    import math
+    from functools import reduce
+    from operator import mul
+
+    # use a double lambda to bind reduce and mul right now
+    math.prod = (lambda fun, op: lambda iterable: fun(op, iterable, 1))(reduce, mul)
+
+    del typing, math, mul, reduce
+
+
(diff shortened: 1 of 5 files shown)


scico/__init__.py
```diff
@@ -12,6 +12,27 @@
 
 import sys
 
+# Python 3.7 compatibility
```
bwohlberg, line 15: Instead of the `try/except` dance, perhaps it would be cleaner to just directly check `sys.version_info`?

Also, perhaps put all the 3.7 compatibility stuff into a separate file (e.g. `_py37.py`) to be imported by `__init__.py`?

scico/numpy/util.py
```diff
@@ -18,6 +19,11 @@
 from ._blockarray import BlockArray
 
 
+def prod(iterable):
```
bwohlberg, line 22: Add a comment here that this can be replaced by `from math import prod` in 3.8+.

scico/_python37.py
```diff
@@ -0,0 +1,24 @@
...
+# package.
+
+""" Alter the behavior of (monkey patch) several modules so that SCICO
+is compatible with Python 3.7, allowing it to run on Google Colab.
+This file should be removed when/if Colab upgrades to Python 3.8.
+"""
+
+import sys
+
+if sys.version_info.minor == 7:
```
bwohlberg, line 14: Perhaps safer to also check that major version is 3 to avoid incorrect logic when 4.7 comes around?

PR #302, opened 2022-05-23
Resolve import error

bwohlberg:
Running the tests manually on the command line results in the error
```
>       if snp.util.is_complex_dtype(x0.dtype):
E       AttributeError: module 'scico.numpy' has no attribute 'util'
```
This is resolved by importing `util.py` in `scico/numpy/__init__.py`.

It isn't clear why this did not trigger an error in the CI tests.

### DIFF
--- a/scico/numpy/__init__.py
+++ b/scico/numpy/__init__.py
@@ -21,7 +21,7 @@ import numpy as np
 
 import jax.numpy as jnp
 
-from . import _wrappers
+from . import _wrappers, util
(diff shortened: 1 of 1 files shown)


Michael-T-McCann (approved):
I checked this on my system: the test fails when run alone (`pytest scico/test/test_solver.py`) but not when the entire suite is run (`pytest`). I suspect this is down to how imports work in pytest, but I don't know the details.

PR #304, opened 2022-05-23
Clean up `ray.tune` interface

bwohlberg:
Clean up `ray.tune` interface and example scripts.

Also add some utility scripts for running `pytest` in various contexts.


bwohlberg:
The minor drop in coverage is on code that is difficult to test, and will not be addressed.

Michael-T-McCann (approved):

Issue #276, opened 2022-04-13
Test failures with GPUs


bwohlberg:
Two unit tests on `main` are failing when run on GPUs. One of them should be easy to resolve via the usual strategy of changing a comparison tolerance, but the other one looks a bit more serious.

```
=================================== FAILURES ===================================
__________________________ test_minimize[CG-float32] ___________________________

dtype = <class 'jax.numpy.float32'>, method = 'CG'

    @pytest.mark.parametrize("dtype", [snp.float32, snp.complex64])
    @pytest.mark.parametrize("method", ["CG", "L-BFGS-B"])
    def test_minimize(dtype, method):
	from scipy.linalg import block_diag

	B, M, N = (4, 3, 2)

	# Models a 12x8 block-diagonal matrix with 4x3 blocks
	A, key = random.randn((B, M, N), dtype=dtype)
	x, key = random.randn((B, N), dtype=dtype)
	y = snp.sum(A * x[:, None], axis=2)  # contract along the N axis

	# result by directly inverting the dense matrix
	A_mat = block_diag(*A)
>       expected = np.linalg.pinv(A_mat) @ y.ravel()

scico/test/test_solver.py:172:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
<__array_function__ internals>:5: in pinv
    ???
/miniconda3/envs/py39gpu/lib/python3.9/site-packages/numpy/linalg/linalg.py:2002: in pinv
    u, s, vt = svd(a, full_matrices=False, hermitian=hermitian)
<__array_function__ internals>:5: in svd
    ???
/miniconda3/envs/py39gpu/lib/python3.9/site-packages/numpy/linalg/linalg.py:1660: in svd
    u, s, vh = gufunc(a, signature=signature, extobj=extobj)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

err = 'invalid value', flag = 8

    def _raise_linalgerror_svd_nonconvergence(err, flag):
>       raise LinAlgError("SVD did not converge")
E       numpy.linalg.LinAlgError: SVD did not converge

/miniconda3/envs/py39gpu/lib/python3.9/site-packages/numpy/linalg/linalg.py:97: LinAlgError
[...]

PR #318, opened 2022-07-28
Bump test tolerance

Michael-T-McCann:
Addresses the half of #276 that I could reproduce.


bwohlberg (approved):
I haven't run the tests with a GPU, but I assume you have.

PR #322, opened 2022-08-03
Add PDHG solver with support for non-linear operators

bwohlberg:
Add PDHG solver with support for non-linear operators.

### DIFF
--- a/examples/scripts/denoise_cplx_tv_pdhg.py
+++ b/examples/scripts/denoise_cplx_tv_pdhg.py
@@ -10,7 +10,7 @@ Complex Total Variation Denoising (ADMM)
 
 This example demonstrates solution of a problem of the form
 
-  $$\argmin_{\mathbf{x}} \; f(\mathbf{x}) + g(C \mathbf{x}) \;,$$
+  $$\argmin_{\mathbf{x}} \; f(\mathbf{x}) + g(C(\mathbf{x})) \;,$$
 
 where $C$ is a nonlinear operator, via non-linear PDHG
 :cite:`valkonen-2014-primal`. The example problem represents total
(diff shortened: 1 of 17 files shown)


examples/scripts/denoise_cplx_tv_pdhg.py
```diff
@@ -0,0 +1,221 @@
...
+"""
+σ = 0.25  # noise standard deviation
+noise, key = scico.random.randn(x_gt.shape, seed=1, dtype=snp.complex64)
+y = x_gt + σ * noise
+
+
+"""
+Denoise with standard total variation.
+"""
+λ_tv = 6e-2
```
Michael-T-McCann, line 67: The result for standard TV looks undersmoothed. Can this be increased? Or is the problem that somehow the Re and Im part _are_ smooth, but the mag/phase representation still looks noisy?

examples/scripts/denoise_l1tv_admm.py
```diff
@@ -37,7 +37,7 @@
 """
 N = 256  # image size
 phantom = SiemensStar(16)
-x_gt = snp.pad(discrete_phantom(phantom, 240), 8)
+x_gt = snp.pad(discrete_phantom(phantom, N - 16), 8)
```
Michael-T-McCann, line 40: 👍

Michael-T-McCann (changes requested):

examples/scripts/denoise_cplx_tv_pdhg.py
bwohlberg: The magnitude is smooth and the phase is noise. Pushing up the regularization improves the magnitude, but it smooths the phase more, which drops the SNR. I didn't do an automated parameter search, but I hand tuned to what I believe is pretty close to the minimum MSE setting.

examples/scripts/denoise_cplx_tv_pdhg.py
```diff
@@ -0,0 +1,221 @@
...
+"""
+Create a ground truth image.
+"""
+N = 256  # image size
+phantom = SiemensStar(16)
+x_mag = snp.pad(discrete_phantom(phantom, N - 16), 8) + 1.0
+x_mag /= x_mag.max()
+# Create reference image with structured magnitude and random phase
+x_gt = x_mag * snp.exp(-1j * scico.random.randn(x_mag.shape, seed=0)[0])
+x_gt = jax.device_put(x_gt)  # convert to jax type, push to device
```
bwohlberg, line 53: @Michael-T-McCann: Is this really necessary? Since `x_gt` is created by an `snp` function, it's already a `DeviceArray`.

scico/_generic_operators.py
```diff
@@ -273,12 +273,33 @@ def jvp(self, primals, tangents):
 
         return jax.jvp(self, primals, tangents)
 
+    def jhvp(self, *primals):
```
bwohlberg, line 276: @Michael-T-McCann: Thoughts on the name of the method? I'm open to suggestions if you think this one is not so great.

examples/scripts/denoise_cplx_tv_pdhg.py
Michael-T-McCann: Agreed, remove it.

scico/_generic_operators.py
Michael-T-McCann: I don't mind it given that jax has settled on `jvp` and `vjp`.

Michael-T-McCann (approved):

PR #329, opened 2022-08-31
Clean up `Operator` and `LinearOperator` module structure

bwohlberg:
Clean up `Operator` and `LinearOperator` module structure.


scico/linop/_misc.py.BAK
```diff
@@ -0,0 +1,219 @@
+# Copyright (C) 2020-2022 by SCICO Developers
```
Michael-T-McCann, line 1: This file meant to be here?

Michael-T-McCann (approved):

scico/linop/_misc.py.BAK
bwohlberg: No! Good catch, thanks. Not sure how that sneaked in there.

Issue #342, opened 2022-09-16
Tests generate large number of numpy warnings


bwohlberg:
Running the tests under `numpy` 1.22.4 generates many warnings

```UserWarning: The value of the smallest subnormal for <class 'numpy.float32'> type is zero.```

It seems that these errors are due to one of the dependencies setting a compiler flag that results in [undesirable handling of subnormal floating point values](https://moyix.blogspot.com/2022/09/someones-been-messing-with-my-subnormals.html). The most likely candidates are `bm3d` and `bm4d`, since these errors disappear if the tests are run without these two packages installed. It's worth noting, though, that these packages don't seem to generate the errors when used independently of their `scico` interfaces, so it's possible that there is something problematic in these interfaces rather than in the packages themselves.

**Comment by bwohlberg** [comment]
Confirmed using the test recommended in the linked analysis that the dynamic libraries included with the `bm3d` package do mess with the CPU floating point flags, as also described there.

**Comment by bwohlberg** [comment]
An additional update on this issue. The dynamic libraries distributed with `bm3d` and `bm4d` were tested using a minor variation of the test script recommended in the link above:
```python
import sys

from ctypes import CDLL

CDLL(sys.argv[1])

import numpy as np

print("%e" % np.finfo(np.float32).smallest_subnormal)
print("Smallest subnormal is zero: ", np.finfo(np.float32).smallest_subnormal == 0.0)
```
which was run as in
```bash
python fptest.py /[full path redacted]/lib/python3.9/site-packages/bm4d/libbm4d.so
```

All of the dynamic libraries distributed with `bm3d` and `bm4d` test positive for the undesirable floating point behaviour, but strangely, when testing by actually importing the two packages, only `bm4d` seems to result in the undesirable handling of subnormal numbers.

bwohlberg:
This issue is confirmed under Linux, but does not seem to be present in the MacOS version of `bm4d`.

bwohlberg:
Resolved with the recent `bm3d` and `bm4d` updates.

PR #343, opened 2022-09-19
Mitigate impact of #342

bwohlberg:
Mitigate impact of #342.

### DIFF
--- a/scico/denoiser.py
+++ b/scico/denoiser.py
@@ -5,7 +5,26 @@
 # user license can be found in the 'LICENSE' file distributed with the
 # package.
 
-"""Denoisers."""
+"""
+Interfaces to standard denoisers.
+
+**Warning**: The :func:`bm3d` and :func:`bm4d` functions are
+implemented as interfaces to the `bm3d
+<https://pypi.org/project/bm3d>`__ and `bm4d
+<https://pypi.org/project/bm4d>`__ packages respectively. The current
+versions of these packages (bm3d 3.0.9 and bm4d 4.0.0) appear to
+be compiled with compiler options that `change the behavior of the
+floating point unit
+<https://moyix.blogspot.com/2022/09/someones-been-messing-with-my-subnormals.html>`__
+in a way that results in lower numerical accuracy, and that persist
+for the duration of the process that loads them. If either of these
+packages are installed, simply loading this module
+(:mod:`scico.denoiser`), is sufficient to inflict this loss of
+numerical precision on all other calculations run within the same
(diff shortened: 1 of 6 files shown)


scico/denoiser.py
```diff
@@ -5,7 +5,26 @@
 # user license can be found in the 'LICENSE' file distributed with the
 # package.
 
-"""Denoisers."""
+"""
```
tbalke, line 8: The user sees this when looking carefully at the source code or with `help(scico.denoiser)`. Do you think this is enough? Perhaps if these packages are installed there could be an additional warning at the time of import?

tbalke (approved):

scico/denoiser.py
bwohlberg: The user should also see this when looking at the online API docs for `scico.denoisers`. Good question as to whether it's enough. A warning on importing the module is not a bad idea.

scico/denoiser.py
```diff
@@ -8,22 +8,18 @@
...
-numerical precision on all other calculations run within the same
-process, even if neither of :func:`bm3d` or :func:`bm4d` are actually
-used. Users who are not making use of either :func:`bm3d` or
-:func:`bm4d` are advised not to install the corresponding packages.
+for the duration of the process that loads it. If this package is
+installed, simply loading this module (:mod:`scico.denoiser`), is
+sufficient to inflict this loss of numerical precision on all other
+calculations run within the same process, even if :func:`bm4d` is not
+actually used. Users who are not making use of :func:`bm4d` are
+advised not to install the corresponding package. For additional information, see `scico issue #342 <https://github.com/lanl/scico/issues/342>`__.
```
tbalke, line 22: line too long

bwohlberg: Fixed.

scico/denoiser.py
```diff
@@ -25,6 +41,19 @@
...
+    if hasattr(finfo, "smallest_subnormal"):  # can't test with older numpy versions
+        if finfo.smallest_subnormal == 0.0:
+            warnings.warn(
+                "Importing module bm4d has had an adverse affect on floating point "
+                "accuracy. See the documentation for module scico.denoiser, and "
+                "scico issue #342."
+            )
+            warnings.filterwarnings(
+                action="ignore",
+                message="^The value of the smallest subnormal",
```
tbalke, line 56: is this caret character there on purpose?

tbalke (approved):

scico/denoiser.py
bwohlberg: Yes, `message` is a regex.

PR #353, opened 2022-10-17
Indicator documentation notation

tbalke:
Make class docstrings agree with prox docstrings


tbalke:
Equivalently we could change the prox docstrings

bwohlberg (approved):

bwohlberg:
There is still a `>=0` subscript for the `I` in the equation for the prox of `NonNegativeIndicator`.

Issue #240, opened 2022-03-03
Solver for least square problems


Michael-T-McCann:
Currently, solving a simple least squares problem, min_x || Ax - b ||, in SCICO requires calling conjugate gradient on ATA and ATb. For a user without much inverse problem experience, this may not be intuitive. Let's add a solver like `numpy.linalg.lstsq` that takes A and b as input.

Easy approach: just make a wrapper around the existing CG.

PR #354, opened 2022-10-20
Resolve issue #240

bwohlberg:
Resolve issue #240.


Michael-T-McCann (approved):
Great, and thanks for the auto-`x0`, too.

PR #372, opened 2022-12-21
Make compatible with jax_0.4.1

Michael-T-McCann:
* `isinstance(x, DeviceArray)` needed to be changed to `isinstance(x, jnp.ndarray)`, which returns `True` for all kinds of jax arrays, including abstract ones, which I think is what we usually want. For more info, see https://github.com/google/jax/issues/2115.

* Needed to change the `BlockArray` function wrapping because it could not handle positional-only arguments.

* NOT fully addressed: jax has changed from `DeviceArray` to `Array` as its default array type, see https://github.com/google/jax/blob/39b14b1b1fc021ec45e431bebd53ac19a9111789/CHANGELOG.md. Probably much of our documentation talks about `DeviceArray`s still.

### DIFF
--- a/scico/numpy/util.py
+++ b/scico/numpy/util.py
@@ -9,8 +9,6 @@ from typing import Any, List, Optional, Tuple, Union
 import numpy as np
 
 import jax
-from jax.interpreters.pxla import ShardedDeviceArray
-from jax.interpreters.xla import DeviceArray
 
 import scico.numpy as snp
 from scico.typing import ArrayIndex, Axes, AxisIndex, BlockShape, DType, JaxArray, Shape
@@ -50,15 +48,6 @@ def ensure_on_device(
                 f"Will cast it to DeviceArray. "
                 f"To suppress this warning cast all np.ndarrays to DeviceArray first.",
                 stacklevel=2,
-            )
-
-        elif not isinstance(
-            array,
-            (DeviceArray, BlockArray, ShardedDeviceArray),
-        ):
-            raise TypeError(
-                "Each element of parameter arrays must be ndarray, DeviceArray, BlockArray, or "
-                f"ShardedDeviceArray; Argument {i+1} of {len(arrays)} is {type(arrays[i])}."
             )
 
         arrays[i] = jax.device_put(arrays[i])
--- a/docs/source/include/blockarray.rst
+++ b/docs/source/include/blockarray.rst
@@ -32,8 +32,8 @@ appropriate. For example,
     ((2, 3), (3,))
 
     >>> x * 2  # returns BlockArray   # doctest: +ELLIPSIS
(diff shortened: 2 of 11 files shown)


scico/numpy/util.py
```diff
@@ -52,15 +50,6 @@ def ensure_on_device(
                 stacklevel=2,
             )
 
-        elif not isinstance(
```
Michael-T-McCann, line 55: This might be controversial. My logic: instead of listing the types we are okay with, implicitly be okay with any type that can have `device_put` called on it.

scico/numpy/util.py
```diff
@@ -52,15 +50,6 @@ def ensure_on_device(
...
-            array,
-            (DeviceArray, BlockArray, ShardedDeviceArray),
-        ):
-            raise TypeError(
-                "Each element of parameter arrays must be ndarray, DeviceArray, BlockArray, or "
-                f"ShardedDeviceArray; Argument {i+1} of {len(arrays)} is {type(arrays[i])}."
-            )
-
         arrays[i] = jax.device_put(arrays[i])
```
Michael-T-McCann, line 53: This might be controversial. My logic: instead of listing the types we are okay with, implicitly be okay with any type that can have `device_put` called on it.

bwohlberg:
> NOT fully addressed: jax has changed from DeviceArray to Array as its default array type, see https://github.com/google/jax/blob/39b14b1b1fc021ec45e431bebd53ac19a9111789/CHANGELOG.md. Probably much of our documentation talks about DeviceArrays still.

Please open an issue as a reminder that this still needs to be taken care of.

scico/numpy/util.py
bwohlberg: This seems reasonable to me.

bwohlberg:
> [Skip some doctests to make compatible with both old and new scico](https://github.com/lanl/scico/pull/372/commits/a149f0d0bb85937ba3bce7f9b16ed655b86e7ecd)

Why is the doctest skipping necessary? Does jax provide some sort of mechanism for backward compatibility on this transition from `DeviceArray` to `jnp.ndarray`?

The unit tests are failing, apparently due to remaining doctest issues.

Michael-T-McCann:
> Why is the doctest skipping necessary?

In old jax, commands return things like `DeviceArray([[ 2,  6, 14], [ 4,  4,  2]], dtype=int64)` while in new jax that's `Array([[ 2,  6, 14], [ 4,  4,  2]], dtype=int64)`. I can sometimes tell doctest that both are okay using `...`, but that doesn't seem to work at the start of a line.

Michael-T-McCann:
> Please open an issue as a reminder that this still needs to be taken care of.

Done, see #373.

PR #375, opened 2023-01-05
Create common base class for optimizer classes.

bwohlberg:
Create common base class for optimizer classes. A significant fraction of the iterations stats initialization code is moved to the base class.

In addition, fix some faulty logic in the determination of whether an objective function can be evaluated.


scico/optimize/_pgm.py
```diff
@@ -226,41 +163,21 @@ def __init__(
...
         r"""
-
         Args:
             f: Loss or Functional object with `grad` defined.
             g: Instance of Functional with defined prox method.
             L0: Initial estimate of Lipschitz constant of f.
             x0: Starting point for :math:`\mb{x}`.
             step_size: helper :class:`StepSize` to estimate the Lipschitz
                 constant of f.
-            maxiter: Maximum number of PGM iterations to perform.
```
Michael-T-McCann, line 241: I miss seeing `maxiter` listed in the docs for each optimizer. What do you think about putting it back (while still refactoring out `itstat_options`)? Or maybe mention it in the help string for `**kwargs`?

bwohlberg: I see arguments for an against.

For: It would improve readability of individual optimizer documentation, and is a relatively small change.

Against: One of the reasons for use of `**kwargs` and moving the docs for those options to the base class is to avoid duplication of docs, which inevitably leads to errors when changes are made. Keeping the `maxiter` docs in the derived classes would be inconsistent with this principle.

Thoughts?

Michael-T-McCann: It's a tough call. I value the readability of the docs over the maintenance issue, but not strongly. And, the PR already is as it is, so perhaps we leave it.

bwohlberg: Let's go with the PR as-is. We can make the suggested changes later if we decide it's a serious issue.

PR #390, opened 2023-04-08
Reduce github `pytest` workflow runtime using `pytest-split`

bwohlberg:
Reduce github `pytest` workflow runtime using `pytest-split`.


.github/workflows/lint.yml
```diff
@@ -2,33 +2,27 @@
...
     steps:
-      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
         with:
-          python-version: "3.8"
-
+          python-version: "3.9"
```
Michael-T-McCann, line 18: The docs state "SCICO requires Python version 3.7 or later." Given that, I suggest that it would be safer to run the tests on 3.7, as I assume code that works in 3.7 will also work in 3.9 but probably not the reverse.

.github/workflows/mypy.yml
```diff
@@ -2,32 +2,25 @@
...
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
           submodules: recursive
       - name: Install Python 3
-        uses: actions/setup-python@v1
+        uses: actions/setup-python@v4
         with:
-          python-version: 3.8
+          python-version: "3.9"
```
Michael-T-McCann, line 23: See previous comment on python version.

.github/workflows/pypi_upload.yml
```diff
@@ -15,13 +15,13 @@ jobs:
...
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
           submodules: recursive
       - name: Install Python 3
-        uses: actions/setup-python@v1
+        uses: actions/setup-python@v4
         with:
-          python-version: 3.8
+          python-version: "3.9"
```
Michael-T-McCann, line 24: version

.github/workflows/pytest_latest.yml
```diff
@@ -13,19 +13,15 @@ on:
...
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
           submodules: recursive
       - name: Install Python 3
-        uses: actions/setup-python@v1
+        uses: actions/setup-python@v4
         with:
-          python-version: 3.8
+          python-version: "3.9"
```
Michael-T-McCann, line 24: version

.github/workflows/pytest_ubuntu.yml
```diff
@@ -44,15 +35,15 @@ jobs:
             miniforge-version: latest
             activate-environment: test-env
             use-mamba: true
-            python-version: 3.9
+            python-version: "3.9"
```
Michael-T-McCann, line 38: version

.github/workflows/pytest_ubuntu.yml
```diff
@@ -81,22 +72,42 @@ jobs:
...
+
+  coverage:
+    needs: test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python 3.9
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.9"
```
Michael-T-McCann, line 97: version

.github/workflows/test_examples.yml
```diff
@@ -41,10 +33,10 @@ jobs:
             miniforge-version: latest
             activate-environment: test-env
             use-mamba: true
-            python-version: 3.9
+            python-version: "3.9"
```
Michael-T-McCann, line 36: version

Michael-T-McCann (commented):
Cool! One thought on Python version (see comment) otherwise good to go.

.github/workflows/lint.yml
bwohlberg: I'm not sure I agree. I would argue that it's more important to test against the "recommended" version of Python than the minimal supported version. It's perhaps a bit more likely that code that works in 3.7 will work in 3.9 than vice-versa, but it's far more serious if it doesn't work in 3.9 since 3.7 is only supported to allow use of google colab. Perhaps a reasonable alternative would be to introduce a periodic (weekly or montly) test across multiple Python version, along the same lines as the existing "jax latest" test.

Issue #400, opened 2023-04-13
Numerous warnings with jaxlib/jax >= 0.4.3


bwohlberg:
Using `scico` with `jaxlib` / `jax` >= 0.4.3 results in numerous  warnings, such as
```
scico/numpy/_wrappers.py:61
scico/scico/numpy/_wrappers.py:61: UserWarning: In call to wrap_recursively, name spacing is not in target_dict
    warnings.warn(f"In call to wrap_recursively, name {name} is not in target_dict")
```

PR #401, opened 2023-04-13
Address #400

bwohlberg:
Address #400.


scico/numpy/_wrapped_function_lists.py
```diff
@@ -1,3 +1,14 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2022-2023 by SCICO Developers
+# All rights reserved. BSD 3-clause License.
+# This file is part of the SPORCO package. Details of the copyright
+# and user license can be found in the 'LICENSE.txt' file distributed
+# with the package.
+
+
+"""Lists of functions to be wrapped in scico.numpy."""
```
Michael-T-McCann, line 9: some additional explanatory text might help:
"These are intended to be the functions in jax.numpy (https://jax.readthedocs.io/en/latest/jax.numpy.html) that should either (1) map over the blocks of a block array (for math functions); (2) map over a tuple of tuples to create a block array (for creation functions); or (3) reduce a block array to a scalar (for reductions).

The links to the numpy docs in the comments are useful for telling these three cases apart, but include extra functions that are not in jax.numpy and should not be listed here.
"

bwohlberg: Docs updated.

PR #405, opened 2023-04-14
Replace generic `Exception` with more specific exceptions where possible

bwohlberg:
Replace generic `Exception` with more specific exceptions where possible. Add pre-commit and CI tests for use of `raises Exception`.


Michael-T-McCann (approved):
Fine with me.

Issue #331, opened 2022-09-06
Error when applying the adjoint of a `DiagonalStack` of `TomographicProjector`


Michael-T-McCann:
The following code produces and error
```python
import numpy as np
from scico.linop.radon_astra import TomographicProjector
from scico.linop import DiagonalStack
import scico.numpy as snp

N = 10
H = DiagonalStack([TomographicProjector((N, N), 1.0, N, np.linspace(0, np.pi, N))])
H.T @ snp.zeros(H.output_shape)
```

```python
NotImplementedError: Transpose rule (for reverse-mode differentiation) for 'custom_vjp_call_jaxpr' not implemented
```

The error goes away if `jit=False` is passed to `DiagonalStack`. But, oddly, I can't reproduce it using a `TomographicProjector` alone without the stack. May relate to #323.

PR #414, opened 2023-05-08
Make adjoint and gram get auto set separately 

Michael-T-McCann:
Previously, when a LinOp needed _either_  a gram operator _or_ an adjoint, both would be automatically generated. In particular, this meant that user-defined adjoint would be overwritten if the user did not also set the gram. This PR splits the two.

Addresses #331


bwohlberg (approved):
Looks good. I took the liberty of bumping a test threshold to address the CI failure on OSX.

PR #415, opened 2023-05-09
Resolve #165

### LINKED ISSUE #165: Make `scico.DType` type more specific
Right now, arguments of type `scico.DType` can be anything:
  https://github.com/lanl/scico/blob/445412ef1c4bfb017628f13b35cf8972e51b872f/scico/typing.py#L29

It would be more useful for automated type checking to set this to the types we intend to support. As a first guess, I'd say `Union[np.float32, np.float64, np.complex64]`.

bwohlberg:
Resolve #165 and modify Sphinx configuration to avoid expansion of the much longer definition of `DType` from making parts of the docs unreadable. Also introduce new `TypeAlias` declaration in `scico.typing` and add a support function for use in future example scripts.


docs/source/conf.py
```diff
@@ -352,6 +360,46 @@ def patched_parse(self):
...
+#   https://github.com/tox-dev/sphinx-autodoc-typehints/issues/284
+#   https://github.com/tox-dev/sphinx-autodoc-typehints/blob/main/README.md
+def typehints_formatter_function(annotation, config):
+    markup = {
+        DType: ":obj:`~scico.typing.DType`",
+        Optional[DType]: ":obj:`~typing.Optional`\ [\ :obj:`~scico.typing.DType`\ ]",
+        Union[DType, Sequence[DType]]: (
+            ":obj:`~typing.Union`\ [\ :obj:`~scico.typing.DType`\ , "
+            ":obj:`~typing.Sequence`\ [\ :obj:`~scico.typing.DType`\ ]]"
+        ),
```
Michael-T-McCann, line 390: As I understand it, you have to add each type that uses `DType` here. Correct? Perhaps add a comment along the lines of "add new types here".

Michael-T-McCann (approved):

docs/source/conf.py
bwohlberg: Yes, correct. Good point -- will do.

PR #419, opened 2023-05-15
Add broadcast_block_shapes

### LINKED ISSUE #146: Implement `snp.broadcast_shapes`
There are times when we would like to know the result of broadcasting a `BlockArray` against another `BlockArray`. Note that the [docs](https://scico.readthedocs.io/en/latest/classes.html#blockarray-class) should already define how this broadcasting will work. 

Currently, we do this with https://github.com/lanl/scico/blob/5ae48c4f826af1c98e72eb90b207d5c5f3c68b60/scico/linop/_linop.py#L196
Ideally, we could do this by implementing `snp.broadcast_shapes` so that it works correctly with nested shapes. What makes this hard is that `scico.numpy` (1) has some complex metaprogramming and (2) circular import problems that stop you from accessing `utils.is_nested`.

For more context, see #144.

Michael-T-McCann:
Resolves #146.


scico/numpy/util.py
```diff
@@ -47,7 +47,6 @@ def ensure_on_device(
     arrays = list(arrays)
 
     for i, array in enumerate(arrays):
-
         if isinstance(array, np.ndarray):
             warnings.warn(
                 f"Argument {i+1} of {len(arrays)} is a numpyy ndarray. "
```
bwohlberg, line 52: Might as well fix the old typo in `numpyy`.

scico/numpy/util.py
```diff
@@ -222,6 +221,36 @@ def is_nested(x: Any) -> bool:
...
+    Args:
+        shape_a: First array shape.
+        shape_b: Second array shape.
+
+    Returns:
+        A (possibly nested) shape tuple.
+
+    Example:
+        >>> broadcast_nested_shapes(((1, 1, 3), (2, 3, 4)), ((3,), (2, 1, 4)))
+        ((1, 1, 3), (2, 3, 4))
```
bwohlberg, line 239: Perhaps choose an example that doesn't just give a result of the same shape as the first argument?

scico/numpy/util.py
```diff
@@ -222,6 +221,36 @@ def is_nested(x: Any) -> bool:
     return isinstance(x, (list, tuple)) and any([isinstance(_, (list, tuple)) for _ in x])
 
 
+def broadcast_nested_shapes(
+    shape_a: Union[Shape, BlockShape], shape_b: Union[Shape, BlockShape]
+) -> Union[Shape, BlockShape]:
+    r"""Compute the result of applying a broadcasting binary operator to
+    (block) arrays with (possibly nested) shapes `shape_a` and `shape_b`.
+
```
bwohlberg, line 229: 1. First line of docstring should be a one-line summary for the function index.
2. It may be worth explicitly mentioning that this is an extension of `numpy.broadcast_shapes` (with a cross-reference to that function) to include nested shapes corresponding to the shape of a `BlockArray`.

bwohlberg (approved):
Approved subject to merge of #421.

PR #429, opened 2023-06-29
Cristina/docs flax

crstngc:


bwohlberg (approved):

bwohlberg:
Looks good, but perhaps it would be worth making the content of the modules in `scico.flax.train` accessible directly from `scico.flax.train`, i.e. rename the submodules to have an initial `_` so that they aren't documented themselves?

PR #433, opened 2023-07-28
Add a jax-based X-ray projector

Michael-T-McCann:
For now, it's the simplest (2D, not very configurable) example. Extension to 3D slice-by-slice should be straightforward.

The structure of the code requires input. I designed this so that the projector could easily be cut and pasted outside of SCICO if desired.

Timing versus ASTRA is encouraging, see `ct_projector_comparison.py`.

### DIFF
--- a/scico/linop/_xray.py
+++ b/scico/linop/_xray.py
@@ -12,6 +12,7 @@ X-ray projector classes.
 from functools import partial
 
 import numpy as np
+
 import jax
 import jax.numpy as jnp
 from jax.typing import ArrayLike
@@ -87,7 +88,7 @@ class ParallelFixedAxis2dProjector:
 
             return inds
 
-        inds = compute_inds(angles)
+        inds = compute_inds(angles)  # (len(angles), *im_shape)
 
         @partial(jax.vmap, in_axes=(None, 0))
         def project_inds(im: ArrayLike, inds: ArrayLike) -> ArrayLike:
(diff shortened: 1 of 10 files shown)


scico/linop/_xray.py
```diff
@@ -0,0 +1,125 @@
...
+import jax
+import jax.numpy as jnp
+from jax.typing import ArrayLike
+
+from scico.typing import Shape
+
+from ._linop import LinearOperator
+
+
+class XRayProject(LinearOperator):
```
bwohlberg, line 26: Naming is not consistent with existing CT projector classes: worth discussion.

scico/linop/_xray.py
```diff
@@ -0,0 +1,125 @@
...
+        """
+        self._eval = projector.project
+
+        super().__init__(
+            input_shape=projector.im_shape,
+            output_shape=(len(projector.angles), *projector.det_shape),
+        )
+
+
+class ParallelFixedAxis2dProjector:
```
bwohlberg, line 48: Quite a few different properties squeezed into the name! If any of these may be generalized (e.g. 2d or 3d), consider moving the property out of the name and making it a parameter when the alternative is also supported.

scico/linop/_xray.py
```diff
@@ -0,0 +1,125 @@
...
+    """Parallel ray, single axis, 2D X-ray projector."""
+
+    def __init__(
+        self,
+        im_shape: Shape,
+        angles: ArrayLike,
+        det_length: Optional[int] = None,
+        do_dithering: bool = True,
```
bwohlberg, line 56: How about just `dither`?

scico/linop/_xray.py
Michael-T-McCann: discussion: call all of them radon transform, clarify in the docs

Michael-T-McCann: New discussion: `XRayTransform`

PR #456, opened 2023-10-05
Added a new Functional for TV Norm 

shnaqvi:
implementing its proximal operator using the fast subiteration-free algorithm proposed by [Kamilov, 2016](https://doi.org/10.1109/TIP.2016.2629449). 

P.S. Tested it on a simple deblur problem with TV regularization using synthetic image generated with anisotropic gaussian blur and the reconstructed result is promising, with 10 iterations returning in 11 sec on 4kx3k image without GPU acceleration. 

image


bwohlberg:
I'll check this carefully as soon as I can find the time, but for now, can you try to track down why some of the tests are failing?

shnaqvi:
> can you try to track down why some of the tests are failing?
I've fixed all but 1 error, because of which it's failing the same test for MacOS and Ubuntu. My `TV2DNorm` has the same output shape as the input which is also the case for an already defined norm `L1MinusL2Norm`, but `TV2DNorm` keeps failing that test with `ValueError` that `Cannot stack LinearOperators with nested output shapes`. Do you have a suggestion on why could this be?

bwohlberg:
I will try to find some time to figure out what's causing the error. Could you give me permission to push to your fork/branch?

shnaqvi:
> Could you give me permission to push to your fork/branch?

Just did!

bwohlberg:
Received, thanks.

bwohlberg:
The test error was a result of the `TV2DNorm` being tested with `BlockArray` input, which it doesn't support. I've disabled that, but it still needs to be tested.

bwohlberg:
The cross-section of the solution in the plot in the initial PR comment looks suspicious: it's piece-wise constant at the bottom, but not at the top. It would be worth checking whether this due to a bug in the code.

shnaqvi:
> It would be worth checking whether this due to a bug in the code.

I see what you mean @bwohlberg. But I'm not able to see what could be causing it. I've gone through Kamilov's algorithm again in detail and am not seeing any obvious issues in my straightforward 50-line implementation of his proposed proximal of the TV Norm 🤔. I've also sent another email to Kamilov today to see if he can review my reference implementation. I get better results after 30 iterations, but there's still some overshoot in the top of the cross-section profile.
image

shnaqvi:
@bwohlberg , I heard back from Ulugbek Kamilov who couldn't share a reference implementation but didn't point out any issues with this implementation either. He said and I can verify that the results with this TV regularizer are converging to piecewise constant in the limit of iterations getting larger. Please let me know if you have any other reservations.

bwohlberg:
See the attached test script -- it looks as if the prox doesn't actually do anything.

[tvnormtest.zip](https://github.com/lanl/scico/files/12908090/tvnormtest.zip)

scico/functional/_norm.py
```diff
@@ -477,3 +479,110 @@ def prox(
...
+
+    has_eval = True
+    has_prox = True
+
+    def __init__(self, tau: float = 1.0):
+        r"""
+        Args:
+            tau: Parameter :math:`\tau` in the norm definition.
+        """
+        self.tau = tau
```
bwohlberg, line 508: Does this play a useful role? The base `Functional` already handles multiplication by a scalar.

scico/functional/_norm.py
```diff
@@ -477,3 +479,110 @@ def prox(
         svdU, svdS, svdV = snp.linalg.svd(v, full_matrices=False)
         svdS = snp.maximum(0, svdS - lam)
         return svdU @ snp.diag(svdS) @ svdV
+
+
+class TV2DNorm(Functional):
```
bwohlberg, line 484: It would be better for this to be able to handle arbitrary numbers of dimensions.

scico/functional/_norm.py
```diff
@@ -477,3 +479,110 @@ def prox(
...
+    For a :math:`M \times N` matrix, :math:`\mb{A}`, by default,
+
+    .. math::
+           \norm{\mb{A}}_{\text{TV}} = \sum_{n=1}^N \sum_{m=1}^M
+           \abs{\nabla{A}_{m,n}} \;.
+
+    The proximal operator of this norm is currently only defined for 2
+    dimensional data.
+
+    For `BlockArray` inputs, the TV norm follows the reduction rules
```
bwohlberg, line 496: Since there doesn't seems to be an obvious way to generalize to `BlockArray` inputs, this should be removed.

shnaqvi: Sure

scico/functional/_norm.py
shnaqvi: @bwohlberg , we can extend it to 3D by adding another set of conditional statements if axis == 2. However, it doesn't look straightforward to extend it to arbitrary number of dimensions because we need to make in-place updates and in `Jax` we do this using `arr.at().set()` framework. But while numpy allows for indexing using `.take()` function, `Jax` doesn't seem to allow it. Do you know of a way to update immutable arrays in Jax by indexing using equivalent of `take()` to replace `at().set()`?

scico/functional/_norm.py
shnaqvi: Sorry I didn't understand. Do you mean the role of the `tau` parameter? It affects the threshold in the `prox` function.

shnaqvi:
> See the attached test script -- it looks as if the prox doesn't actually do anything.
> [tvnormtest.zip](https://github.com/lanl/scico/files/12908090/tvnormtest.zip)

You should set the parameter to the `prox` function smaller than `1e-2`. Anything larger starts to clip off the haar coefficients to the point that they are zeroed out. I have found it helpful to print out the following two quantities inside the `prox` function call to guide the setting of the parameter passed to the `prox` function.
```
print(f'haar mean: {snp.mean(snp.abs(self.ht2(v, axis=ax, shift=False)))}')
print(f'thresh: {thresh}') 
```

That said, I didn't quote understand the role of the TV operator in this simple example. As in, for this 1D signal, what's the interpretation of the prox. function applied to it?

scico/functional/_norm.py
bwohlberg: But it just multiplies the prox scaling parameter, no? So what does it add over that parameter?

bwohlberg:
> That said, I didn't quite understand the role of the TV operator in this simple example. As in, for this 1D signal, what's the interpretation of the prox. function applied to it?

It's just a simple test to see that the prox does to the signal. I would expect that it has the same effect as TV denoising, but it seems to do nothing. I tried a smaller prox scaling (less than 1e-2) but that also has no effect (and one expects a bigger effect for a bigger scaling, not the other way around).

scico/functional/_norm.py
bwohlberg: Let's come back to this after having resolved the other issues.

scico/functional/_norm.py
shnaqvi: @bwohlberg , sure it does scales the prox parameter in the `prox()` function, but it is also used in the `__call__` function to evaluate the Norm to scale the output, right?

shnaqvi:
Well it does have an effect on my end. Here is what I'm seeing. As the scaling parameter is incrementally increased from `5e-3` to `2e-1`, the signal keeps getting increasingly suppressed. That's what you'd expect right?

image

shnaqvi:
Looking at the implementation, do you see the code deviating from the [Kamilov, 2016](https://doi.org/10.1109/TIP.2016.2629449) algorithm? It would be great if we can try to integrate it this week, so we can start using it from `scico`.

bwohlberg:
> Well it does have an effect on my end. Here is what I'm seeing. As the scaling parameter is incrementally increased from `5e-3` to `2e-1`, the signal keeps getting increasingly suppressed. That's what you'd expect right?

That's certainly closer to what I would expect than I was seeing, which was no effect at all. (I would really expect to see the smooth waveform being clipped to something closer to a square wave.) Could you share the script you used?

bwohlberg:
> Looking at the implementation, do you see the code deviating from the [Kamilov, 2016](https://doi.org/10.1109/TIP.2016.2629449) algorithm in any way? It would be great if we can try to integrate it this week, so we can start using it from `scico`.

I'm not convinced that the order of the 2D transform and shrinkage is correct, which is a consequence of the 1D forward transform and shrinkage being fused into a single function. The correct approach would be to compute the full 2D transform, shrink the highpass coefficients, and then apply the pseudo-inverse.

It would also be much cleaner (and easy to extend to arbitrary numbers of dimensions) if the highpass and lowpass coefficients were computed by convolution with the [1, -1] and [1, 1] filters (the representations here omit the necessary normalization).

shnaqvi:
> Could you share the script you used?

I used this same script ([tvnormtest.py.zip](https://github.com/lanl/scico/files/13221966/tvnormtest.py.zip)) that you shared for plotting. And this is the `_norm.py` ([_norm.py.zip](https://github.com/lanl/scico/files/13221973/_norm.py.zip))
 with `TV2DNorm` implementation that the test script used.

bwohlberg:
> I used this same script ([tvnormtest.py.zip](https://github.com/lanl/scico/files/13221966/tvnormtest.py.zip)) that you shared for plotting. And this is the `_norm.py` ([_norm.py.zip](https://github.com/lanl/scico/files/13221973/_norm.py.zip)) with `TV2DNorm` implementation that the test script used.

The `TV2DNorm` implementation is not the same as in the PR branch?

shnaqvi:
> I'm not convinced that the order of the 2D transform and shrinkage is correct, which is a consequence of the 1D forward transform and shrinkage being fused into a single function. The correct approach would be to compute the full 2D transform, shrink the highpass coefficients, and then apply the pseudo-inverse.

@bwohlberg , that is what I was initially doing which gave me better results. Yes we can totally do that. Instead of shrinking just the coefficients from the difference operator in the Haar transform, apply the shrinkage operator to the entire forward transform output. And I did check how the TV Norm does in our simple test case of a sinusoid and results are visually the same.

bwohlberg:
> @bwohlberg , that is what I was initially doing which gave me better results. Yes we can totally do that. Instead of shrinking just the coefficients from the difference operator in the Haar transform, apply the shrinkage operator to the entire forward transform output. And I did check how the TV Norm does in our simple test case of a sinusoid and results are visually the same.

That's not quite what I meant. It seems pretty clear from the paper that the shrinkage should only be applied to the highpass coefficients, but this should be done *after* the decomposition has been computed in both spatial directions.

shnaqvi:
> The `TV2DNorm` implementation is not the same as in the PR branch?

Sorry @bwohlberg, I went back investigating. Indeed, this is different (attached again here, [_norm.py.zip](https://github.com/lanl/scico/files/13222327/_norm.py.zip)). And it is basically the original version of the implementation that I had submitted in the PR, which has the shrinkage operator applied to the entirety of the forward transform. 

I verified that with the shrinkage operator applied inside the forward transform function, I am able to reproduce your result, which is that I am not seeing any effect of the proximal parameter. But I am not able to spot any coding errors that could be causing this behavior. 

So basically with the shrinkage operator applied on the entirety of the forward transform, we are getting expected clipping behavior and none otherwise.

shnaqvi:
> > @bwohlberg , that is what I was initially doing which gave me better results. Yes we can totally do that. Instead of shrinking just the coefficients from the difference operator in the Haar transform, apply the shrinkage operator to the entire forward transform output. And I did check how the TV Norm does in our simple test case of a sinusoid and results are visually the same.
> 
> It seems pretty clear from the paper that the shrinkage should only be applied to the highpass coefficients, but this should be done _after_ the decomposition has been computed in both spatial directions.

I am seeing it differently. If you please refer to equation 20b in [Kamilov, 2016](https://www.merl.com/publications/docs/TR2017-002.pdf), the shrinkage operator is applied inside the summation sign after the `kth` decomposition is computed. And later we sum over both the spatial dimensions and the shifts. Do you agree?

bwohlberg:
> I am seeing it differently. If you please refer to equation 20b in [Kamilov, 2016](https://www.merl.com/publications/docs/TR2017-002.pdf), the shrinkage operator is applied inside the summation sign after the `kth` decomposition is computed. And later we sum over both the spatial dimensions and the shifts. Do you agree?

I agree. My concern was that the decompositions were being mixed along different axes with interleaved shrinkage, but on looking again I see that I was not reading the code correctly. So I don't see any obvious errors in the implementation.

shnaqvi:
> (I would really expect to see the smooth waveform being clipped to something closer to a square wave.)

@bwohlberg , since TV norm is minimizing the summation of the local difference in the image, I think we'll see the features in the image become more smooth (in the sense of high frequencies suppressed, while preserving the low frequencies, so as to not introduce newer sharp edges).

Here, I did some more testing with an image having both low and high frequency features (sinusoidal and square boundaries). I also compared the two implementations of the proximal operator of the TV norm, 1) where we apply shrinkage to just the difference operator or high pass features (labeled "Shrink-Highpass") and 2) where we apply the shrinkage to the entirety of the forward Haar transform (labeled "Shrink-Forward").

Since in my original implementation ("Shrink-Forward"), we were shrinking the entire transform, we begin to see clipping of the signal and the dynamic range is lost, whereas "Shrink-Highpass" retains the dynamic range while minimizing the integral of the image gradient magnitude (see below).
image

And as the proximal parameter increases, "Shrink-Highpass"'s proximal function returns a more smooth signal (to a certain limit), while "Shrink-Forward" returns a more clipped signal (to an absolute zero in the limit).  (see below).
image

> So I don't see any obvious errors in the implementation.

So if you're okay, can we merge the PR to the master? :-)

bwohlberg:
>[...]
> And as the proximal parameter increases, "Shrink-Highpass"'s proximal function returns a more smooth signal (to a certain limit), while "Shrink-Forward" returns a more clipped signal (to an absolute zero in the limit). (see below). 
> [...]

Thanks for adding the additional results. It seems clear that the shrinkage should only be applied to the highpass component.

> > So I don't see any obvious errors in the implementation.
> 
> So if you're okay, can we merge the PR to the master? :-)

We still need tests for the new code. I am working on that, as well as an alternative implementation that supports input arrays of arbitrary dimensionality.

shnaqvi:
> Thanks for adding the additional results. It seems clear that the shrinkage should only be applied to the highpass component.

@bwohlberg, for completeness, I further denoised over 100 iterations of APGM with identity forward operator and this TV Prior (Shrink-Highpass) and the results show a convincingly denoised image, amid heavy gaussian noise. 
image

image

bwohlberg:
Please take a look at the new pull request in your fork of `scico`.

bwohlberg:
@shnaqvi: Merging shortly. Thanks for your help in getting this useful feature added.

shnaqvi:
Pleasure is mine, and thanks for proactively optimizing and integrating it! Next up, Proximal Averaged APGM for composite priors minimization! 😃

PR #457, opened 2023-10-06
Rename linear problem solver and add parameter type checking

bwohlberg:
Rename linear problem solver and add parameter type checking.

### DIFF
--- a/scico/solver.py
+++ b/scico/solver.py
@@ -579,7 +579,7 @@ def golden(
 
 
 class MatrixATADSolver:
-    r"""Solver for linear system involving a symmetric product plus a diagonal.
+    r"""Solver for linear system involving a symmetric product.
 
     Solve a linear system of the form
 
--- a/scico/solver.py
+++ b/scico/solver.py
@@ -709,7 +709,7 @@ class MatrixATADSolver:
                 :class:`Diagonal`, specifies the diagonal elements
                 of :math:`D`.
             W: Matrix :math:`W`. Specifies the diagonal elements of
-                :math:`W`. Defaults to ones.
+                :math:`W`. Defaults to an array with unit entries.
             cho_factor: Flag indicating whether to use Cholesky
                 (``True``) or LU (``False``) factorization.
             lower: Flag indicating whether lower (``True``) or upper
@@ -727,7 +727,7 @@ class MatrixATADSolver:
         else:
             D = jnp.array(D)
             if not D.ndim in [1, 2]:
-                raise ValueError("If matrix, D should be 1D or 2D.")
+                raise ValueError("If MatrixOperator, D should be 1D or 2D.")
 
         if W is None:
             W = snp.ones(A.shape[0], dtype=A.dtype)
--- a/scico/solver.py
+++ b/scico/solver.py
@@ -54,6 +54,7 @@ import numpy as np
 
 import jax
 import jax.experimental.host_callback as hcb
+import jax.numpy as jnp
 import jax.scipy.linalg as jsl
 
 import scico.numpy as snp
@@ -260,7 +261,7 @@ def minimize(
 
 def minimize_scalar(
     func: Callable,
-    bracket: Optional[Union[Sequence[float]]] = None,
+    bracket: Optional[Sequence[float]] = None,
     bounds: Optional[Sequence[float]] = None,
     args: Union[Tuple, Tuple[Any]] = (),
     method: str = "brent",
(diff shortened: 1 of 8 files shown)


scico/solver.py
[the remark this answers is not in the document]
Michael-T-McCann: For the one-liner, how about something like "Solver to invert A^* A + D where A is a convolution and D is diagonal"? I think that gives the whole picture and remains short enough.

scico/solver.py
```diff
@@ -579,7 +578,7 @@ def golden(
     return r
 
 
-class ATADSolver:
+class MatrixATADSolver:
     r"""Solver for linear system involving a symmetric product plus a diagonal.
```
Michael-T-McCann, line 582: Similar to above, maybe something like "Solver to invert A^* W A + D, where..."

scico/solver.py
```diff
@@ -710,14 +715,27 @@ def __init__(
         """
         if isinstance(A, MatrixOperator):
             A = A.to_array()
+        elif not isinstance(A, Array):
```
Michael-T-McCann, line 718: I believe that if we implemented `__array__` in `MatrixOperator`, then `jnp.array(A)` would work for either MatrixOperators or JAX arrays. The input dtype could be `array_like`. Similar for `D`. Might be cleaner/less coupled to `MatrixOperator` if written that way.

scico/solver.py
```diff
@@ -596,12 +595,18 @@ class ATADSolver:
...
-    :math:`D \in \mbb{R}^{N \times N}`. The solution is computed by
-    factorization of matrix :math:`A^T W A + D` and solution via Gaussian
-    elimination. If :math:`D` is diagonal and :math:`N < M` (i.e.
-    :math:`A W A^T` is smaller than :math:`A^T W A`), then
-    :math:`A W A^T + D` is factorized and the original problem is solved
-    via the Woodbury matrix identity
+    :math:`D \in \mbb{R}^{N \times N}`. :math:`A` must be an instance of
+    :class:`.MatrixOperator` or an array; :math:`D` must be an instance
+    of :class:`.MatrixOperator`, :class:`.Diagonal`, or an array, and
+    :math:`W`, if specified, must be an instance of :class:`.Diagonal`
```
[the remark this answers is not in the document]
bwohlberg: It seems difficult to do this effectively without discussing the types.

scico/solver.py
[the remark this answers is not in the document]
bwohlberg: > In `ConvATADSolver`, it says ":math:`D` is a circular convolution operator". Is that correct or should it be a diagonal? Because the short docstring says "sum of convolutions plus diagonal linear system"

Good catch. The one-liner is correct. Main text fixed.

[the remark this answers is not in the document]
bwohlberg: > For the one-liner, how about something like "Solver to invert A^* A + D where A is a convolution and D is diagonal"? I think that gives the whole picture and remains short enough.

It's a bit more complicated than that: `A` itself is not just a convolution. Including the full details would seem to necessitate using more than the single line available, but I'm open to further suggestions.

scico/solver.py
```diff
@@ -596,12 +595,18 @@ class ATADSolver:
 
     where :math:`A \in \mbb{R}^{M \times N}`,
     :math:`W \in \mbb{R}^{M \times M}` and
```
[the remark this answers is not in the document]
bwohlberg: Addressed.

scico/solver.py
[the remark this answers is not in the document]
bwohlberg: Modified. There is a complication that's too involved to address in the one-liner: the function works for any matrix `D`, but the efficient solution via factorization is only partially available if `D` is not diagonal.

bwohlberg: Same issue as above: it's not straightforward to actually fit this into a single line.

scico/solver.py
bwohlberg: This seems like a good suggestion. Having not worked with `__array__` myself, perhaps you could help with this?

Michael-T-McCann: I have addressed this in a commit. Please check, especially the new text in the arguments section of the docs.

bwohlberg: Looks good.

scico/solver.py
```diff
@@ -727,7 +727,7 @@ def __init__(
         else:
             D = jnp.array(D)
             if not D.ndim in [1, 2]:
-                raise ValueError("If matrix, D should be 1D or 2D.")
+                raise ValueError("If MatrixOperator, D should be 1D or 2D.")
```
Michael-T-McCann, line 730: This error can be raised by either an array or a MatrixOperator, maybe should be "If array or MatrixOperator..."

bwohlberg: Fixed.

PR #459, opened 2023-10-10
Resolve #445

### LINKED ISSUE #445: Linear operator multiplication with singleton array not supported
This issue is intended as a distillation of the bug component of #442, which also addresses some broader questions.

In its essence, the problem is that:
- some `scico` functions, such as `linop.operator_norm`, return singleton arrays rather than scalar floats (this behaviour is ubiquitous due to the behaviour of `jax.numpy` ufuncs, e.g. `jax.numpy.sqrt(4.0)` returns an array, not a float)
- in certain contexts, such as when such a computed value is passed as the `lam` parameter of `loss.SquaredL2Loss.prox`, it is multiplied by a linear operator within the internal calculations.

A quick fix would be to cast `lam` to a float within `loss.SquaredL2Loss.prox`, but this does not address the broader problem that such issues are equally likely to arise in other contexts.

A more general solution would be to add support for multiplication by singleton arrays to linear operators, but this additional flexibility is not currently reflected in typing annotations, and properly accounting for it is likely to require significant changes. 

An alternative approach would be to take a careful look at functions nominally expected to return scalars, explicitly casting to float where necessary. A component of this would be to adopt a coherent policy on when to use `numpy` functions instead of `jax.numpy` functions since the latter will invariably return a singleton array. Again, this would require comprehensive changes.

bwohlberg:
Resolve #445.


scico/operator/_operator.py
```diff
@@ -47,7 +49,11 @@ def _wrap_mul_div_scalar(func: Callable) -> Callable:
 
     @wraps(func)
     def wrapper(a, b):
-        if np.isscalar(b) or isinstance(b, jax.core.Tracer):
```
Michael-T-McCann, line 50: I think there should be a utility function for checking if something is a SCICO scalar so we can keep the definition consistent. This can't be the only place this comes up. Would also allow us to document and test it.

scico/operator/_operator.py
```diff
@@ -47,7 +49,11 @@ def _wrap_mul_div_scalar(func: Callable) -> Callable:
...
+        if (
+            snp.isscalar(b)
+            or isinstance(b, jax.core.Tracer)
+            or (isinstance(b, jaxlib.xla_extension.ArrayImpl) and b.size == 1)
```
Michael-T-McCann, line 55: I don't think `jax.core` or `jaxlib` should be involved. `(isinstance(b, jax.Array)` should work for arrays and tracers.

Michael-T-McCann, line 55: I prefer `b.ndim==0` to `b.size==1`.

Michael-T-McCann (changes requested):

scico/operator/_operator.py
bwohlberg: Agreed. See new function `scico.numpy.is_scalar_equiv`. (I expect you may take issue with the abbreviation in the name.)

scico/operator/_operator.py
bwohlberg: This is an interesting one. In principle an array with unit size and dimension can also function as a scalar in multiplication and division. If `ndim` is 0 the pseudo-scalar `dtype` does not affect the result `dtype`, but if `ndim` is 1 it does, which greatly complicates the implementation, so I agree that it's better to base the test on `ndim` to exclude this case.

scico/test/test_numpy_util.py
```diff
@@ -192,3 +193,10 @@ def test_broadcast_nested_shapes():
         (1, 2, 3),
         (1, 7, 4, 3),
     )
+
+
+def test_is_scalar_equiv():
+    assert is_scalar_equiv(1e0)
+    assert is_scalar_equiv(snp.array(1e0))
```
Michael-T-McCann, line 200: throw in `assert is_scalar_equiv(snp.sum(snp.zeros(1)))` just to directly address the original cause of all this?

scico/operator/_operator.py
Michael-T-McCann: According to https://jax.readthedocs.io/en/latest/jax.typing.html#jax-typing-best-practices `jax.Array` is currently "annotation for any JAX array or tracer" so `Tracer` is redundant. It probably wasn't when we first implemented this.

scico/test/test_numpy_util.py
bwohlberg: Good idea. Done.

bwohlberg:
> couple suggestions

Done.

Michael-T-McCann (approved):

PR #463, opened 2023-10-31
Remove `ensure_on_device` function

### LINKED ISSUE #451: Reevaluate need for `ensure_on_device` function
Currently, the `ensure_on_device` function from scico/numpy/util.py interferes with deployment of SCICO with jax.distributed utils. This seems due mainly to the call to `jax.device_put` function. When distributing computations between different hosts, it is important to be able to partition data under different schemes and let jax.distributed to control it. Any outside definition of how data should be partitioned would be superseded by the local call to `jax.device_put`, which completely ignores the runtime context. A cheap fix is to remove the `jax.device_put` call, but perhaps, it would be better to completely remove the `ensure_on_device` function. Besides this, it is also pending to consider the interplay between `BlockArrays` and jax.distributed.

bwohlberg:
Remove `ensure_on_device` function. Resolves #451.

### DIFF
--- a/examples/scripts/ct_abel_tv_admm_tune.py
+++ b/examples/scripts/ct_abel_tv_admm_tune.py
@@ -14,14 +14,14 @@ parameters for the companion [example script](ct_abel_tv_admm.rst). The
 `ray.tune` class API is used in this example.
 
 This script is hard-coded to run on CPU only to avoid the large number of
-warnings that are emitted when GPU resources are requested but not available,
-and due to the difficulty of supressing these warnings in a way that does
-not force use of the CPU only. To enable GPU usage, comment out the
-`os.environ` statements near the beginning of the script, and change the
-value of the "gpu" entry in the `resources` dict from 0 to 1. Note that
-two environment variables are set to suppress the warnings because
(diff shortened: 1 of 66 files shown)


examples/scripts/ct_abel_tv_admm_tune.py
```diff
@@ -82,8 +81,8 @@ def setup(self, config, x_gt, x0, y):
         this case). The remaining parameters are objects that are passed
         to the evaluation function via the ray object store.
         """
-        # Put main arrays on jax device.
-        self.x_gt, self.x0, self.y = jax.device_put([x_gt, x0, y])
```
Michael-T-McCann, line 86: Why remove `device_put` (here and in other scripts)? I read that as "put these matrices on my GPU if I have one" and that seems perfectly reasonable (and orthogonal to the `ensure_on_device` issue).

bwohlberg: It was replaced with conversion to a jax array using `jnp.array` or `snp.array` in the other scripts since I think that more closely matches what we want (ensure it's a jax array rather than more explicitly put it on a "device"), but this specific example may have been overealous.

Michael-T-McCann (approved):

PR #466, opened 2023-11-08
Added a new solver for Proximal Averaged Projected Gradient Method

shnaqvi:
Added a new solver for solving composite prior minimization problem, called Accelerated Proximal Averaged Projected Gradient Method. The optimization problem is given by:

$\text{argmin}_x f(x) + \bar{g}(x),$
$\  \text{where}\ \  \bar{g}(x) = \sum_i \rho_i g_i(x),$
$\  \text{and}\ \  \rho_i \ge 0, \ \sum_i \rho_i = 1$

This addresses the enhancement request #456


shnaqvi:
**Note**: 

1. New solver has been added as a standalone class deriving from the parent `Optimize` class. I couldn't make it derive from the existing `PGM` class without modifying this other class also, because it combines both the gradient descent and projection steps in one line but for `proximal-averaged` method, we need to do these steps separately. 
2. I couldn't figure out how to use several `PGMStepSize` routines with this new solver, so I haven't exposed the feature to dynamically evaluate the Lipschitz constant at runtime. It needs to be provided beforehand.

bwohlberg:
Closing PR since alternative solution merged.

PR #467, opened 2023-11-09
Resolve obstacles to jitting of some functionals

bwohlberg:
Resolve obstacles to jitting of some functionals and add corresponding tests.


Michael-T-McCann (approved):

bwohlberg:
For the record, there was a discussion of the apparent lack of support for jitting of static methods: google/jax#7702.

PR #471, opened 2023-11-15
Fix fractional centers in CircularConvolve

Michael-T-McCann:
Using a fractional filter center for a `CircularConvolve` operator currently gives nonsense results. This PR adds a test for this nonsense (checking that the imaginary part of the result of filtering is small) and changes the way centers are handled to fix the problem. Relevant references:

F. Candocia and J. C. Principe, "Comments on "Sinc interpolation of discrete periodic signals," in IEEE Transactions on Signal Processing, vol. 46, no. 7, pp. 2044-2047, July 1998, doi: 10.1109/78.700979.

"S. -C. Pei and Y. -C. Lai, "Closed Form Variable Fractional Time Delay Using FFT," in IEEE Signal Processing Letters, vol. 19, no. 5, pp. 299-302, May 2012, doi: 10.1109/LSP.2012.2191280.

### DIFF
--- a/scico/test/linop/test_circconv.py
+++ b/scico/test/linop/test_circconv.py
@@ -160,10 +160,11 @@ class TestCircularConvolve:
 
     def test_fractional_center(self):
         """A fractional center should keep outputs real"""
-        x, _ = uniform(minval=-1, maxval=1, shape=(3, 4), key=self.key)
+        x, _ = uniform(minval=-1, maxval=1, shape=(4, 5), key=self.key)
         h, _ = uniform(minval=-1, maxval=1, shape=(2, 2), key=self.key)
         A = CircularConvolve(h=h, input_shape=x.shape, h_center=[0.1, 2.7])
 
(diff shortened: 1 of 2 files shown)


scico/test/linop/test_circconv.py
```diff
@@ -160,6 +158,21 @@ def test_center(self, center):
             shift = -center[0]
         np.testing.assert_allclose(A @ x, snp.roll(B @ x, shift), atol=1e-5)
 
+    def test_fractional_center(self):
+        """A fractional center should keep outputs real"""
+        x, _ = uniform(minval=-1, maxval=1, shape=(4, 5), key=self.key)
+        h, _ = uniform(minval=-1, maxval=1, shape=(2, 2), key=self.key)
```
bwohlberg, line 164: It would be a bit better to not discard the key returned by the first call to `unform`, instead using it as the second key.

bwohlberg (approved):

PR #480, opened 2023-12-08
Resolve #479

### LINKED ISSUE #479: Scaling of `FiniteDifference` linear operator broken
Scaling of the `FiniteDifference` linear operator broken; this simple example throws an exception:
```python
from scico.linop import FiniteDifference

A = FiniteDifference(input_shape=(8, 8))
B = 2.0 * A
```
The problem is due to `FiniteDifference` being derived from `VerticalStack`, and the use of `self.__class__` in the implementation of `__mul__` in `VerticalStack`.

bwohlberg:
Resolve #479.

The solution is to simply remove the `VerticalStack`-specific arithmetic operator definitions, the need for which is unclear.

### DIFF
--- a/scico/linop/_stack.py
+++ b/scico/linop/_stack.py
@@ -9,8 +9,6 @@
 
 from __future__ import annotations
 
-import operator
-from functools import partial
 from typing import Optional, Sequence, Union
 
 import scico.numpy as snp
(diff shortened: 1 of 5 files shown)


Michael-T-McCann (approved):
With this PR, we lose the ability to add two `VerticalStack`s together and get another `VerticalStack` rather than a generic `Operator` or `LinearOperator`. I am willing to give that up to reduce code duplication. There may be other places in the code with a similar pattern.

bwohlberg:
> With this PR, we lose the ability to add two `VerticalStack`s together and get another `VerticalStack` rather than a generic `Operator` or `LinearOperator`. I am willing to give that up to reduce code duplication. There may be other places in the code with a similar pattern.

This was intentional. It isn't clear what real benefit there is to this capability, and it requires considerable additional complexity. It seems simplest to remove it for now, and re-enable it if we discover some significant use for it.

PR #484, opened 2023-12-14
Resolve #483

### LINKED ISSUE #483: Incorrect handling of functions with no return value in `scico.numpy`
The wrapper functions used in constructing `scico.numpy` from `numpy` do not correctly handle functions without any return value, such as `numpy.testing.assert_allclose`:
```python
import scico.numpy as snp
from scico.numpy.testing import assert_allclose

x = snp.array([1.0, 2.0])
y = snp.array([1.0, 2.0])
r = assert_allclose(x, y)
# As expected, function that doesn't return anything has a None return value
print(r is None)

xb = snp.blockarray((x, x))
yb = snp.blockarray((y, y))
r = assert_allclose(xb, yb)
# A BlockArray is constructed from the implicit None return value,
print(r)
```
This is presumably a result of a bug/oversight in `map_func_over_blocks`.

bwohlberg:
Resolve #483.

### DIFF
--- a/scico/numpy/_wrappers.py
+++ b/scico/numpy/_wrappers.py
@@ -96,6 +96,32 @@ def map_func_over_tuple_of_tuples(func: Callable, map_arg_name: Optional[str] = 
     return mapped
 
 
+def _num_blocks_in_args(*args, **kwargs):
+    """Count the number of BlockArray arguments."""
+    first_ba_arg = next((arg for arg in args if isinstance(arg, BlockArray)), None)
+    if first_ba_arg is None:
+        first_ba_kwarg = next((v for k, v in kwargs.items() if isinstance(v, BlockArray)), None)
+        if first_ba_kwarg is None:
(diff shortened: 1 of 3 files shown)


scico/numpy/_wrappers.py
```diff
@@ -96,34 +96,64 @@ def mapped(*args, **kwargs):
...
+def map_void_func_over_blocks(func):
+    """Wrap a function without a return value so that it maps over all
+    of its BlockArray arguments.
+    """
+
+    @wraps(func)
+    def mapped(*args, **kwargs):
+        num_blocks = _num_blocks_in_args(*args, **kwargs)
+        if num_blocks == 0:
+            func(*args, **kwargs)  # no BlockArray arguments, so no mapping
```
Michael-T-McCann, line 152: I suggest `return` immediately after calling `func` inside this `if`

Michael-T-McCann (approved):

PR #485, opened 2023-12-14
Resolve numerous deprecation warnings

bwohlberg:
Resolve numerous deprecation warnings.


pytest.ini
```diff
@@ -2,4 +2,10 @@
...
 addopts = --doctest-glob="*rst"
 doctest_optionflags = NORMALIZE_WHITESPACE NUMBER
-filterwarnings = ignore::DeprecationWarning:.*.compat
+filterwarnings =
+    ignore::DeprecationWarning:.*pkg_resources.*
+    ignore::DeprecationWarning:.*flax.*
+    ignore::DeprecationWarning:.*.tensorboardx.*
+    ignore::DeprecationWarning:.*xdesign.*
+    ignore:.*pkg_resources.*:DeprecationWarning
+    ignore:.*imp module.*:DeprecationWarning
```
Michael-T-McCann, line 11: Can any of these easily be made more specific to avoid hiding future deprecation warnings from these packages?

Michael-T-McCann (approved):

pytest.ini
bwohlberg: It's not so straightforward, but I can try.

bwohlberg: But I will do that some other time.

PR #487, opened 2023-12-15
Various changes

bwohlberg:
Various changes, including:
* Improvements to the conda environment creation script
* Updates to documentation related to installing a version of `jax` with GPU support
* Replacement of `jax.experimental.host_callback.call` (which is pending deprecation, and no longer seems to function properly on GPU devices) with `jax.pure_callback`
* Address very slow tests for `svmbir` interface
* Bump maximum allowed `jaxlib` and `jax` versions


misc/conda/make_conda_env.sh
```diff
@@ -233,12 +255,10 @@ echo "  conda activate $ENVNM"
...
-echo "Jax installed without GPU support. To avoid warning messages,"
-echo "add the following to your .bashrc or .bash_aliases file"
-echo "  export JAX_PLATFORM_NAME=cpu"
-echo "To include GPU support, reinstall the astra-toolbox conda"
-echo "package on a host with GPUs, and see the instructions at"
-echo "   https://github.com/google/jax#pip-installation-gpu-cuda"
+echo "Jax installed without GPU support. To include GPU support,"
+echo "reinstall the astra-toolbox conda package on a host with GPUs,"
+echo "and see the instructions at"
+echo "   https://jax.readthedocs.io/en/latest/installation.html#nvidia-gpu"
```
Michael-T-McCann, line 261: I don't understand why both Jax and astra-toolbox are mentioned here.

Michael-T-McCann (changes requested):

misc/conda/make_conda_env.sh
bwohlberg: If I recall correctly (see [this issue](https://github.com/astra-toolbox/astra-toolbox/issues/327)), astra is installed without GPU support if installed on a host without GPUs, so the `astra-toolbox` reinstall is necessary to include the required CUDA dependencies.

Michael-T-McCann (approved):

PR #501, opened 2024-02-07
Reduce memory requirements of `Identity` linop

bwohlberg:
Reduce memory requirements of `Identity` linop, and other improvements.

### DIFF
--- a/scico/linop/_circconv.py
+++ b/scico/linop/_circconv.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright (C) 2021-2023 by SCICO Developers
+# Copyright (C) 2021-2024 by SCICO Developers
 # All rights reserved. BSD 3-clause License.
 # This file is part of the SCICO package. Details of the copyright and
 # user license can be found in the 'LICENSE' file distributed with the
@@ -8,8 +8,6 @@
 """Circular convolution linear operators."""
 
 import math
-import operator
-from functools import partial
 from typing import Optional, Sequence, Tuple, Union
 
 import numpy as np
@@ -205,7 +203,7 @@ class CircularConvolve(LinearOperator):
             H_adj_x = H_adj_x.real
         return H_adj_x
 
-    @partial(_wrap_add_sub, op=operator.add)
+    @_wrap_add_sub
     def __add__(self, other):
         if self.ndims != other.ndims:
             raise ValueError(f"Incompatible ndims: {self.ndims} != {other.ndims}.")
@@ -218,7 +216,7 @@ class CircularConvolve(LinearOperator):
             h_is_dft=True,
         )
 
-    @partial(_wrap_add_sub, op=operator.sub)
+    @_wrap_add_sub
     def __sub__(self, other):
         if self.ndims != other.ndims:
             raise ValueError(f"Incompatible ndims: {self.ndims} != {other.ndims}.")
--- a/scico/linop/_circconv.py
+++ b/scico/linop/_circconv.py
@@ -240,10 +240,6 @@ class CircularConvolve(LinearOperator):
         )
 
     @_wrap_mul_div_scalar
-    def __rmul__(self, scalar):
-        return self * scalar
-
-    @_wrap_mul_div_scalar
     def __truediv__(self, scalar):
         return CircularConvolve(
             h=self.h_dft / scalar,
--- a/scico/linop/_diag.py
+++ b/scico/linop/_diag.py
@@ -50,18 +50,18 @@ class Diagonal(LinearOperator):
             input_dtype: `dtype` of input argument. The default,
                ``None``, means `diagonal.dtype`.
         """
-        self.diagonal = diagonal
+        self._diagonal = diagonal
 
         if input_shape is None:
-            input_shape = self.diagonal.shape
+            input_shape = self._diagonal.shape
 
         if input_dtype is None:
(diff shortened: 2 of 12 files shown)


scico/linop/_diag.py
```diff
@@ -156,10 +156,137 @@ def norm(self, ord=None):  # pylint: disable=W0622
...
+            diagonal = scalar * snp.ones((), dtype=input_dtype)
+        super().__init__(
+            diagonal=diagonal,
+            input_shape=input_shape,
+            input_dtype=input_dtype,
+            **kwargs,
+        )
+
+    def _eval(self, x: Union[Array, BlockArray]) -> Union[Array, BlockArray]:
+        return self._diagonal * x
```
Michael-T-McCann, line 190: Not obvious why we need this instead of the `_eval` in `Diagonal`. Maybe a comment to clarify?

scico/linop/_diag.py
```diff
@@ -156,10 +156,137 @@ def norm(self, ord=None):  # pylint: disable=W0622
...
+        **kwargs,
+    ):
+        """
+        Args:
+            scalar: Scaling of the identity.
+            input_shape: Shape of input array.
+            input_dtype: `dtype` of input argument.
+        """
+        if is_nested(input_shape):
+            diagonal = scalar * snp.ones(((),) * len(input_shape), dtype=input_dtype)
```
Michael-T-McCann, line 179: I assume you tried `diagonal = scalar`. I'm curious why that doesn't work.

scico/linop/_diag.py
```diff
@@ -156,10 +156,137 @@ def norm(self, ord=None):  # pylint: disable=W0622
...
+            if isinstance(other, ScaledIdentity):
+                return ScaledIdentity(
+                    scalar=self._diagonal * other._diagonal,
+                    input_shape=self.input_shape,
+                    input_dtype=self.input_dtype,
+                )
+            else:
+                return Diagonal(diagonal=self._diagonal * other.diagonal)
+        else:
+            return self(other)
```
Michael-T-McCann, line 269: Not a problem with this PR, but the previous 50+ lines of code to express `scalar * x` is a code smell. Let's make an issue (or dev note) about looking into these wrappers.

scico/test/linop/test_binop.py
```diff
@@ -0,0 +1,48 @@
...
+    def test_case1(self, operator):
+        A = linop.Convolve(
+            snp.ones((2,)), input_shape=self.input_shape, input_dtype=self.input_dtype, mode="same"
+        )
+        B = linop.Identity(input_shape=self.input_shape, input_dtype=self.input_dtype)
+
+        assert type(A + B) == linop.LinearOperator
+        assert type(B + A) == linop.LinearOperator
+        assert type(2.0 * A + 3.0 * B) == linop.LinearOperator
+        assert type(2.0 * B + 3.0 * A) == linop.LinearOperator
```
Michael-T-McCann, line 24: parameterized by `operator` but `operator` is never used?

scico/test/linop/test_diag.py
```diff
@@ -187,3 +187,146 @@ def test_norm_except(self):
...
+        assert isinstance(sId, linop.ScaledIdentity)
+        snp.testing.assert_allclose(sId @ x, scalar * x, rtol=1e-6)
+
+        Ids = Id * scalar
+        assert isinstance(Ids, linop.ScaledIdentity)
+        snp.testing.assert_allclose(Ids @ x, scalar * x, rtol=1e-6)
+
+        Idds = Id / scalar
+        assert isinstance(Idds, linop.ScaledIdentity)
+        snp.testing.assert_allclose(Idds @ x, x / scalar, rtol=1e-6)
```
Michael-T-McCann, line 332: Another possible issue to create: can we have a standard function to test scaling (and other similar properties) rather than copy/pasting variants of this?

scico/linop/_diag.py
bwohlberg: Inclusion was an oversight. Removed.

scico/linop/_diag.py
bwohlberg: Agreed. Please add a note to #502 if what's already there doesn't fully reflect what you have in mind.

scico/linop/_diag.py
bwohlberg: The initializer for `Diagonal` assumes that `_diagonal` has a `shape` attribute:
https://github.com/lanl/scico/blob/7cef2aa41d0f333e05c940add6ce89199a42a4ea/scico/linop/_diag.py#L56-L59
This could, of course, be modified to check for the scalar case, but that would just entail moving this complexity elsewhere in the code.

scico/test/linop/test_binop.py
bwohlberg: Good catch. Fixed.

scico/test/linop/test_diag.py
bwohlberg: One of many aspects of the tests that could be handled better. But this one shouldn't be terrible to fix. Please make an issue.

Michael-T-McCann (approved):

scico/linop/_circconv.py
```diff
@@ -8,8 +8,6 @@
 """Circular convolution linear operators."""
 
 import math
-import operator
-from functools import partial
```
Michael-T-McCann, line 12: excellent!

Michael-T-McCann (approved):

PR #507, opened 2024-02-28
Add support for approximate isotropic TV norm

bwohlberg:
Add support for approximate isotropic TV norm.


scico/functional/_tvnorm.py
```diff
@@ -1,87 +1,115 @@
...
-    the parameter of the proximal operator. For example, in :class:`.PGM`
-    and :class:`.AcceleratedPGM`, the scaled proximal operator parameter
-    is the regularization parameter divided by the `L0` algorithm
-    parameter, and for :class:`.ADMM`, the scaled proximal operator
-    parameters are the regularization parameters divided by the entries
-    in the `rho_list` algorithm parameter.
+    Generic total variation (TV) norm with approximation of the scaled
+    proximal operator :cite:`kamilov-2016-parallel`
+    :cite:`kamilov-2016-minimizing`.
     """
```
Michael-T-McCann, line 34: I'm not familiar with these papers. How universally applicable is this approximation? If it always works, fine. If not, is it clear from the docs that if you want TV-regularized reconstruction you should _not_ use `TVNorm`? Is it confusing to say that it is "is equivalent to `linop.FiniteDifference` when they are not actually interchangeable?

bwohlberg: The norm is exactly equivalent; it's just the prox that's approximated.

Issue #513, opened 2024-04-24
`RuntimeWarning` when using `ray`


bwohlberg:
The examples and tests that use `ray` are producing warnings
```
RuntimeWarning: os.fork() was called. os.fork() is incompatible with multithreaded code, and JAX is multithreaded, so this will likely lead to a deadlock.
```
This issue and a possible solution are discussed in ray-project/ray#44087.

PR #515, opened 2024-04-25
Rearrange `ray` initialization to avoid `RuntimeWarning`

bwohlberg:
Rearrange `ray` initialization to avoid `RuntimeWarning` resulting from initialization of `ray` after `jax` has been imported, as described in #513.


conftest.py
```diff
@@ -1,21 +1,39 @@
...
 
 import pytest
 
+try:
+    import ray  # noqa: F401
+except ImportError:
+    have_ray = False
+else:
+    have_ray = True
+    ray.init(num_cpus=1)  # required to be here: see ray-project/ray#44087
```
Michael-T-McCann, line 15: Why `num_cpus=1`? I don't see anything about that at ray-project/ray#44087

bwohlberg: It's the location of the `ray.init` call that is "required to be here". The constraint on the number of cpus is so that it will play nicely with tests run in parallel.

bwohlberg: Updated comment in an attempt to clarify this.

Michael-T-McCann (approved):

PR #517, opened 2024-05-09
Add support for `np.newaxis` to indexing syntax supported by `linop.Slice`

bwohlberg:
Add support for `np.newaxis` to indexing syntax supported by `linop.Slice`.

### DIFF
--- a/scico/numpy/util.py
+++ b/scico/numpy/util.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright (C) 2022-2023 by SCICO Developers
+# Copyright (C) 2022-2024 by SCICO Developers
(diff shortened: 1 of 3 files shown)


Michael-T-McCann (approved):
I wish `indexed_shape` could be simpler, but I could not find an alternative. A JIT function that creates an array, indexes it, and returns the shape is attractive, but slices are apparently unhashable (in most Python version). There might be a workaround.

bwohlberg:
> I wish `indexed_shape` could be simpler, but I could not find an alternative.  A JIT function that creates an array, indexes it, and returns the shape is attractive, but slices are apparently unhashable (in most Python version). There might be a workaround.

Feel free to create an issue as a reminder to look into this if you think that's a realistic expectation.

PR #523, opened 2024-05-17
Resolve some deprecation warnings

bwohlberg:
Resolve some deprecation warnings.

### DIFF
--- a/scico/flax/train/traversals.py
+++ b/scico/flax/train/traversals.py
@@ -37,7 +37,7 @@ def clip_positive(params: PyTree, traversal: ModelParamTraversal, minval: float 
         minval: Minimum value to clip selected model parameters and keep
             them in a positive range. Default: 1e-4.
     """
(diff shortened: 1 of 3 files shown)


scico/flax/train/traversals.py
```diff
@@ -37,7 +37,7 @@ def clip_positive(params: PyTree, traversal: ModelParamTraversal, minval: float
         minval: Minimum value to clip selected model parameters and keep
             them in a positive range. Default: 1e-4.
     """
-    params_out = traversal.update(lambda x: jnp.clip(x, a_min=minval), params)
```
Michael-T-McCann, line 40: fine, could also `jnp.clip(x, min=minval)`

Michael-T-McCann (approved):

scico/flax/train/traversals.py
bwohlberg: Unfortunately whether that works is dependent on the jax version in use.

PR #531, opened 2024-06-07
Resolve #530

### LINKED ISSUE #530: Interface to ASTRA FBP broken on GPU devices
The ASTRA FBP interface (`scico.linop.xray.astra.XRayTransform2D.fbp` is broken on GPU devices. This test scripts
```python
import numpy as np
from scico.linop.xray.astra import XRayTransform2D

N = 256
x = np.zeros((N, N))
x[N//4:-N//4, N//4:-N//4] = 1.0

Nproj = 90
angles = np.linspace(0, np.pi, Nproj)
A = XRayTransform2D((N, N), N, 1.0, angles)

y = A(x)
xfbp = A.fbp(y)

print(f"FBP result sum: {np.abs(np.sum(xfbp)):.2e}")
```
indicates a constant zero result when run on a GPU.

bwohlberg:
Resolve #530.

### DIFF
--- a/scico/test/linop/xray/test_astra.py
+++ b/scico/test/linop/xray/test_astra.py
@@ -122,6 +122,11 @@ def test_adjoint_typical_input(testobj):
     adjoint_test(A, x=x, rtol=get_tol())
 
 
+def test_fbp(testobj):
+    x = testobj.A.fbp(testobj.y)
+    assert np.sum(np.abs(x)) > 0.0
+
+
 def test_jit_in_DiagonalStack():
     """See https://github.com/lanl/scico/issues/331"""
     N = 10
--- a/scico/linop/xray/astra.py
+++ b/scico/linop/xray/astra.py
@@ -191,7 +191,6 @@ class XRayTransform2D(LinearOperator):
                <https://www.astra-toolbox.com/docs/algs/FBP_CUDA.html>`__.
         """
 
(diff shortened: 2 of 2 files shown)


scico/test/linop/xray/test_astra.py
```diff
@@ -122,6 +122,11 @@ def test_adjoint_typical_input(testobj):
     adjoint_test(A, x=x, rtol=get_tol())
 
 
+def test_fbp(testobj):
+    x = testobj.A.fbp(testobj.y)
+    assert np.sum(np.abs(x)) > 0.0
```
Michael-T-McCann, line 127: Consider checking that `x` is reasonably close to `testobj.x`.

bwohlberg: That did occur to me, but I decided against it because it wasn't obvious how to choose a reasonable threshold for "reasonably close". But easy to add if you think it's nevertheless worthwhile.

Michael-T-McCann: On further consideration, I think what I'd prefer is a comment to the effect of "checks for bug where `fbp` would return all zeros when because ASTRA CPU FBP algorithm was being called on a GPU system."

crstngc (approved):

scico/test/linop/xray/test_astra.py
bwohlberg: Done. Please confirm approval.

PR #529, opened 2024-06-07
Add JAX 3D X-ray CT projector

Michael-T-McCann:
See example results at https://scico--529.org.readthedocs.build/en/529/examples/ct_projector_comparison_3d.html
Still not clear why the back projections don't look like they match. It seems like calling `.transpose(2, 1, 0)` on the ASTRA back projection gets them closer. My best guess is that this example generates vectors that cause ASTRA back projection to get flipped. I do think SCICO's back projection is correct given that it comes from autograd.

For consideration:
* Is the `XRayTransform(Parallel3dProjector(in_shape, P, out_shape))` syntax too complex? I like that it separates the projector logic from SCICO (makes it easy to cut and paste into other projects), but it's awkward.
    * Removed `XRayTransform`

PR #533, opened 2024-06-12
Improve docs for `linop.xray` subpackage

bwohlberg:
Improve docs for `linop.xray` subpackage. See also lanl/scico-data#49.

The effects of the changes in the built docs may be viewed [here](https://scico--533.org.readthedocs.build/en/533/_autosummary/scico.linop.xray.html#module-scico.linop.xray). These docs are likely to be moved or edited as part of the completion of #529.

### DIFF
--- a/docs/source/conf/20-extensions.py
+++ b/docs/source/conf/20-extensions.py
@@ -11,6 +11,7 @@ extensions = [
     "sphinx.ext.viewcode",
     "sphinxcontrib.bibtex",
     "sphinx.ext.inheritance_diagram",
+    "matplotlib.sphinxext.plot_directive",
     "sphinx.ext.todo",
     "nbsphinx",
 ]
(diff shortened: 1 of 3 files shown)


Michael-T-McCann (commented):
Cool plots!

Gut reaction: I object to the labels "y (axis 0)" and "x (axis 1)" on the SCICO plot. The code is written so that axis 0 goes with x and axis 1 axis with y. The downside of that is that it means rows <-> x and columns <-> y.

We could add the motivation for why SCICO works this way. The idea is to think about rotation (by theta) followed by projection (onto the x-axis). Probably I could improve the `Parallel2dProjector` docs to make this more clear (by avoiding talking about rows and columns).

bwohlberg:
Comments addressed.

Michael-T-McCann (approved):

PR #544, opened 2024-07-19
Improve docs for 3D astra X-ray transform

bwohlberg:
Improve docs for 3D `astra` X-ray transform. Also move existing `matplotlib` figures from the submodule to the main repo.


scico/linop/xray/astra.py
```diff
@@ -227,6 +227,85 @@ class XRayTransform3D(LinearOperator):  # pragma: no cover
...
+
+    .. plot:: pyfigures/xray_3d_vol.py
+       :align: center
+       :include-source: False
+       :show-source-link: False
+
+    The voxels sides have unit length (in arbitrary units), which defines
+    the scale for all other dimensions in the source-volume-detector
+    configuration. Geometry axes `x`, `y`, and `z` correspond to volume
+    array axes 0, 1, and 2 respectively. The projected array axes 0, 1,
```
Michael-T-McCann, line 244: Consider remarking that this is different from astra's default interface, were I think x y z corresponds to 2 1 0.

scico/linop/xray/astra.py
```diff
@@ -227,6 +227,85 @@ class XRayTransform3D(LinearOperator):  # pragma: no cover
...
+       :align: center
+       :include-source: False
+       :show-source-link: False
+
+    Vector :math:`\mb{r}` is not illustrated to avoid cluttering the
+    figure, but will typically be directed toward the center of the
+    detector (i.e. in the direction of :math:`\mb{d}` in the figure.) In
+    practice, since the volume-detector distance does not have a
+    geometric effect for a parallel-beam configuration, :math:`\mb{d}`
+    may be set to the zero vector. Note that the view images must be
```
Michael-T-McCann, line 298: I think that d is useful for handling offsets in the center of rotation, i.e., situations where the center column of a projection view doesn't match up with the rotation axis. Put another way, because it's parallel ray, the component of d perpendicular to the detector doesn't matter, but the other components do matter.

Michael-T-McCann (approved):

scico/linop/xray/astra.py
bwohlberg: Good catch - fixed.

scico/linop/xray/astra.py
bwohlberg: Phrasing modified to avoid over-simplification.

Michael-T-McCann (approved):

PR #547, opened 2024-07-31
Fix docs figure display in furo theme "dark" mode

bwohlberg:
Fix docs figure display in furo theme "dark" mode and improve some figures and their captions.


scico/linop/xray/astra.py
```diff
@@ -252,9 +252,10 @@ class XRayTransform3D(LinearOperator):  # pragma: no cover
        :align: center
        :include-source: False
        :show-source-link: False
-       :caption: Red arrows indicate the direction of the beam towards
-          the detector (orange) and the arrows parallel to the detector
-          indicate the direction of increasing pixel indices.
+       :caption: Each radial arrow indicates the direction of the beam
+          towards the detector (indicated in orange in the "light"
+          display mode) and the arrow parallel to the detector indicates
```
Michael-T-McCann, line 257: While the color inversion in dark mode is cool, might it be simpler to have the figures include a white background? Maybe ugly, but would also this parenthetical statement which I could see being confusing.

Michael-T-McCann (approved):

scico/linop/xray/astra.py
bwohlberg: It would be simpler, but I think the white background would be even uglier than the parenthetical statement. Any suggestions for less confusing phrasing?

PR #551, opened 2024-09-09
Changes to `linop.xray.astra` module

bwohlberg:
Changes to `linop.xray.astra` module:
* Add missing typing
* Fix some docstrings
* Change interface of `convert_to_scico_geometry`

Note:
* Typing still needs some attention. In particular, it's not always clear whether `jax` or `numpy` arrays should be specified, and `jax.ArrayLike` should perhaps be replaced by `numpy.typing.ArrayLike`.
* Some functions do not have any docstrings.

### DIFF
--- a/examples/scripts/ct_projector_comparison_3d.py
+++ b/examples/scripts/ct_projector_comparison_3d.py
@@ -132,7 +132,7 @@ HTy_astra = H_astra.T @ y_astra
 Convert ASTRA geometry to SCICO and project.
 """
 
-P_from_astra = scico.linop.xray.astra.convert_to_scico_geometry(H_astra.vol_geom, H_astra.proj_geom)
+P_from_astra = scico.linop.xray.astra._astra_to_scico_geometry(H_astra.vol_geom, H_astra.proj_geom)
 H_scico_from_astra = XRayTransform(Parallel3dProjector(in_shape, P_from_astra, out_shape))
 
 y_scico_from_astra = H_scico_from_astra @ x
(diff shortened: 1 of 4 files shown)


Michael-T-McCann:
>     * Typing still needs some attention. In particular, it's not always clear whether `jax` or `numpy` arrays should be specified, and `jax.ArrayLike` should perhaps be replaced by `numpy.typing.ArrayLike`.

Putting the specific issue of this module aside, do we have a policy about this for SCICO? For example, `jax.numpy` functions are annotated with `jax._src.typing.ArrayLike -> jax._src.typing.Array` (documented [here](https://github.com/google/jax/blob/main/jax/_src/basearray.py), includes NumPy inputs), whereas `LinearOperator.__call__` is `Union[LinearOperator, Array, BlockArray] -> Union[LinearOperator, Array, BlockArray]` (does not include NumPy inputs).

For this module, it would be convenient if at least JAX inputs were allowed for everything, with internal conversion to NumPy when ASTRA requires it.

Michael-T-McCann (approved):

bwohlberg:
> Putting the specific issue of this module aside, do we have a policy about this for SCICO? For example, `jax.numpy` functions are annotated with `jax._src.typing.ArrayLike -> jax._src.typing.Array` (documented [here](https://github.com/google/jax/blob/main/jax/_src/basearray.py), includes NumPy inputs), whereas `LinearOperator.__call__` is `Union[LinearOperator, Array, BlockArray] -> Union[LinearOperator, Array, BlockArray]` (does not include NumPy inputs).

I don't recall this has ever been discussed in detail. If both jax and numpy array inputs are reasonable, then a Union would make sense. It's often clear, though, what the output type will be if there's an explicit jax.numpy or numpy call in the processing chain.

> For this module, it would be convenient if at least JAX inputs were allowed for everything, with internal conversion to NumPy when ASTRA requires it.

That seems fine.

PR #554, opened 2024-09-19
Add caveat on 3D X-ray projector

Michael-T-McCann:
Add caveat on 3D X-ray projector.

### DIFF
--- a/scico/linop/xray/_xray.py
+++ b/scico/linop/xray/_xray.py
@@ -235,6 +235,11 @@ class XRayTransform2D(LinearOperator):
 class XRayTransform3D(LinearOperator):
     r"""General-purpose, 3D, parallel ray X-ray projector.
 
+    This projector approximates cubical voxels projecting onto
+    rectangular pixels and provides a back projector that is the true
+    adjoint of the forward projector. It is written purely in JAX,
+    allowing it to run on either CPU or GPU and minimizing host copies.
+
     Warning: This class is experimental and may be up to ten times slower
     than :class:`scico.linop.xray.astra.XRayTransform3D`.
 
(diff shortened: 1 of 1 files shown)


scico/linop/xray/_xray.py
```diff
@@ -235,6 +235,9 @@ def _calc_weights(
 class XRayTransform3D(LinearOperator):
     r"""General-purpose, 3D, parallel ray X-ray projector.
 
+    Warning: This class is experimental and may be up to ten times slower
+    than :class:`scico.linop.xray.astra.XRayTransform3D`.
```
bwohlberg, line 239: Perhaps add some positive aspects, e.g. "It is expected to be useful for small problems as, unlike :class:`scico.linop.xray.astra.XRayTransform3D`, it can be run on both CPU and GPU, and the forward and adjoint transforms are differentiable." [The differentiable claim should be confirmed before inclusion!]

Michael-T-McCann: Differentiable wrt to `x`, definitely, but so is `scico.linop.xray.astra.XRayTransform3D` in the way we've wrapped it. I'm not sure about differentiable wrt to the angles. I'd like to call that out of scope for this PR. 

How about 

This projector approximates cubical voxels projecting onto rectangular pixels and provides a back projector that is the true adjoint of the forward projector. It is written purely in JAX, allowing it to run on either CPU or GPU and minimizing host copies.

Warning: This class is experimental and may be up to ten times slower than [scico.linop.xray.astra.XRayTransform3D](https://scico--554.org.readthedocs.build/en/554/_autosummary/scico.linop.xray.astra.html#scico.linop.xray.astra.XRayTransform3D).

PR #555, opened 2024-09-27
Changes required to support `jax` version 0.4.33

bwohlberg:
Changes required to support `jax` version 0.4.33, switch to using the new `orbax-checkpoint`  `CheckpointManager` API, and various other improvements.

### DIFF
--- a/examples/scripts/deconv_odp_train_foam1.py
+++ b/examples/scripts/deconv_odp_train_foam1.py
@@ -66,6 +66,11 @@ os.environ["XLA_FLAGS"] = "--xla_force_host_platform_device_count=8"
 
 import jax
 
+try:
+    from jax.extend.backend import get_backend  # introduced in jax 0.4.33
+except ImportError:
+    from jax.lib.xla_bridge import get_backend
+
 from mpl_toolkits.axes_grid1 import make_axes_locatable
 
 from scico import flax as sflax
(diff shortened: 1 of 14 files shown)


examples/scripts/deconv_odp_train_foam1.py
```diff
@@ -66,6 +66,11 @@
 
 import jax
 
+try:
+    from jax.extend.backend import get_backend  # introduced in jax 0.4.33
+except ImportError:
+    from jax.lib.xla_bridge import get_backend
+
```
Michael-T-McCann, line 73: consider making this a function giving that it is repeated several times

scico/numpy/_blockarray.py
```diff
@@ -174,10 +174,18 @@ def prop_ba(self):
 def _da_method_wrapper(method_name):
     method = getattr(Array, method_name)
 
-    if method.__name__ is None:
+    if method.__name__ is None or method.__qualname__ is None:
         return method
```
Michael-T-McCann, line 178: this is no longer needed given the later logic?

Michael-T-McCann (changes requested):

examples/scripts/deconv_odp_train_foam1.py
bwohlberg: I considered that, but it's going to be removed eventually (once the minimum supported version moves past 0.4.31), so it seemed more appropriate to do this.

scico/numpy/_blockarray.py
bwohlberg: Good point - fixed.

Michael-T-McCann (approved):

PR #557, opened 2024-10-01
Fix `XRayTransform2D` projection dtype and docs

### LINKED ISSUE #542: Incorrect parameter descriptions for `linop.xray.Parallel2dProjector.__init__`
The description of parameters `x0` and `dx` of `linop.xray.Parallel2dProjector.__init__`

https://github.com/lanl/scico/blob/722846da030d8410fbbeaac16ab6e364d49ca1e0/scico/linop/xray/_xray.py#L86-L89

does not match what is implemented

https://github.com/lanl/scico/blob/722846da030d8410fbbeaac16ab6e364d49ca1e0/scico/linop/xray/_xray.py#L99-L100

https://github.com/lanl/scico/blob/722846da030d8410fbbeaac16ab6e364d49ca1e0/scico/linop/xray/_xray.py#L117-L119

Since the default $\sqrt{2}/2$ pixel size differs from those of `astra` and `svmbir`, it would be best to include a clear warning to users in addition to correcting the documentation.

Michael-T-McCann:
resolves #542
hypothesized to resolve a warning about dtypes from the tests

### DIFF
--- a/scico/linop/xray/_xray.py
+++ b/scico/linop/xray/_xray.py
@@ -27,7 +27,7 @@ from .._linop import LinearOperator
 
 
 class XRayTransform2D(LinearOperator):
-    """Parallel ray, single axis, 2D X-ray projector.
+    r"""Parallel ray, single axis, 2D X-ray projector.
 
     This implementation approximates the projection of each rectangular
     pixel as a boxcar function (whereas the exact projection is a
--- a/scico/linop/xray/_xray.py
+++ b/scico/linop/xray/_xray.py
@@ -124,11 +124,19 @@ class XRayTransform2D(LinearOperator):
         )
 
     def project(self, im: ArrayLike) -> snp.Array:
-        """Compute X-ray projection."""
+        """Compute X-ray projection, equivalent to `H @ im`.
+
+        Args:
+            im: Input array representing the image to project.
+        """
         return XRayTransform2D._project(im, self.x0, self.dx, self.y0, self.ny, self.angles)
 
     def back_project(self, y: ArrayLike) -> snp.Array:
-        """Compute X-ray back projection"""
+        """Compute X-ray back projection, equivalent to `H.T @ y`.
+
+        Args:
+            y: Input array representing the sinogram to back project.
+        """
         return XRayTransform2D._back_project(y, self.x0, self.dx, self.nx, self.y0, self.angles)
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


bwohlberg:
Unfortunately the dtype changes thus far don't resolve the warning
```
py312/lib/python3.12/site-packages/jax/_src/ops/scatter.py:92: FutureWarning: scatter inputs have incompatible types: cannot safely cast value from dtype=float64 to dtype=float32 with jax_numpy_dtype_promotion='standard'. In future JAX releases this will result in an error.
```

scico/test/linop/xray/test_xray.py
```diff
@@ -49,7 +49,7 @@ def test_apply():
 def test_apply_adjoint():
     im_shape = (12, 13)
     num_angles = 10
-    x = jnp.ones(im_shape)
+    x = jnp.ones(im_shape, dtype=jnp.float32)
```
bwohlberg, line 52: This is more of a band-aid than a real fix. The `astra` and `svmbir` linop class initializers explicitly force
```python
    input_dtype=np.float32,
    output_dtype=np.float32,
```
when calling `super().__init__`, but neither of the `scico` ones do. They seem to default to `np.float32`, but it would be better to decide whether fixing it to this choice is desirable, and if so, making it explicit, and properly supporting and testing alternatives such as `np.float64` if it is not.

scico/linop/xray/_xray.py
```diff
@@ -146,8 +151,11 @@ def _project(
         # ignored, while inds < 0 wrap around. So we set inds < 0 to ny.
         inds = jnp.where(inds >= 0, inds, ny)
 
+        # avoid incompatible types in the .add (scatter operation)
+        weights = weights.astype(im.dtype)
+
         y = (
-            jnp.zeros((len(angles), ny))
+            jnp.zeros((len(angles), ny), dtype=im.dtype)
```
bwohlberg, line 158: With respect to the longer comment below, this line (and others like it) seem to imply the intent to support different choices of `input_dtype` (and corresponding `output_dtype`).

bwohlberg (approved):

bwohlberg:
I noticed that the public versions of these methods
https://github.com/lanl/scico/blob/c58b8196871f935e2de4fae6f12cf596bf8def14/scico/linop/xray/_xray.py#L126-L132
are not fully document, while the private versions are. It wouldn't hurt to add the missing docs in this PR.

bwohlberg (approved):

PR #561, opened 2024-10-15
Fix XRayTransform boundary conditions

### LINKED ISSUE #560: `XRayTransform2D` fails adjoint test
This example demonstrates a case in which `scico.linop.xray.XRayTransform2D` does not have a properly matched adjoint:
```python
import numpy as np
from scico.linop import valid_adjoint
from scico.linop.xray import XRayTransform2D

N = 256
det_count = int(N * 1.05 / np.sqrt(2.0))
dx = 1.0 / np.sqrt(2)
n_projection = 360
angles = np.linspace(0, np.pi, n_projection, endpoint=False)
A = XRayTransform2D((N, N), angles, det_count=det_count, dx=dx)
print(f"Adjoint test pass: {valid_adjoint(A, A.T, eps=1e-5)}")
print(f"Adjoint accuracy: {valid_adjoint(A, A.T, eps=None):.3e}")
```
The test passes if `det_count = N`.

It would be worth adding a test for this to the tests for this module!

Michael-T-McCann:
addresses #560


bwohlberg (approved):
Consider renaming ` scico/test/linop/xray/test_xray.py` to ` scico/test/linop/xray/test_xray_2d.py`.

PR #568, opened 2024-11-02
Restore running of astra tests

bwohlberg:
Restore running of astra tests, and add some tests for `linop.Convolve`.

### DIFF
--- a/scico/test/linop/xray/test_astra.py
+++ b/scico/test/linop/xray/test_astra.py
@@ -80,6 +80,31 @@ def testobj(request):
     yield XRayTransform2DTest(request.param)
 
 
+def test_init(testobj):
+    with pytest.raises(ValueError):
+        A = XRayTransform2D(
+            input_shape=(16, 16, 16),
+            det_count=16,
+            det_spacing=1.0,
+            angles=np.linspace(0, np.pi, 32, False),
+        )
+    with pytest.raises(ValueError):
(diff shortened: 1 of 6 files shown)


scico/test/linop/xray/test_astra.py
```diff
@@ -30,6 +29,17 @@
 RTOL_GPU_RANDOM_INPUT = 1.0
 
 
+def make_im(Nx, Ny, is_3d=True):
+    x, y = snp.meshgrid(snp.linspace(-1, 1, Nx), snp.linspace(-1, 1, Ny))
```
Michael-T-McCann, line 33: (minor and old bug): The shape of `x` after this line is `(Ny, Nx)` which is inconsistent with how that function is later used (`x = make_im(A.input_shape[0], A.input_shape[1], is_3d=False)`). A quick fix would be ` x, y = snp.meshgrid(snp.linspace(-1, 1, Nx), snp.linspace(-1, 1, Ny), indexing='ij')`.

Michael-T-McCann (changes requested):

scico/test/linop/xray/test_astra.py
bwohlberg: Fixed (also in `test_svmbir`).

Michael-T-McCann (approved):

PR #569, opened 2024-11-06
Add call tracing module

bwohlberg:
Add call tracing module. Also change approach to defining `scico.optimize.PGM.x_step`.


examples/scripts/trace_example.py
```diff
@@ -0,0 +1,109 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# This file is part of the SCICO package. Details of the copyright
+# and user license can be found in the 'LICENSE.txt' file distributed
+# with the package.
+
+r"""
+SCICO Call Tracing
```
Michael-T-McCann, line 8: Should this example appear as a notebook at https://scico--569.org.readthedocs.build/en/569/examples.html? I don't see it there.

examples/scripts/trace_example.py
```diff
@@ -0,0 +1,109 @@
...
+Set up the forward operator and ADMM solver object.
+"""
+lmbda = 1e-1
+A = linop.MatrixOperator(D)
+register_variable(A, "A")
+f = loss.SquaredL2Loss(y=y, A=A)
+g_list = [lmbda * functional.L1Norm(), functional.NonNegativeIndicator()]
+C_list = [linop.Identity((n)), linop.Identity((n))]
+rho_list = [1.0, 1.0]
+maxiter = 10  # number of ADMM iterations
```
Michael-T-McCann, line 82: Consider lowering this to 1 or 2. Would make the output of the script more digestible.

examples/scripts/trace_example.py
```diff
@@ -0,0 +1,109 @@
...
+register_variable(g_list[1], "g_list[1]")
+register_variable(C_list[0], "C_list[0]")
+register_variable(C_list[1], "C_list[1]")
+
+solver = ADMM(
+    f=f,
+    g_list=g_list,
+    C_list=C_list,
+    rho_list=rho_list,
+    x0=A.adj(y),
```
Michael-T-McCann, line 95: When I first looked at this output, I was confused by
```
>> A.adj(Array(32,))
>> Array(128,)
```
because I didn't realize that `Array(128,)` was indicating what was being returned. Would it be simple to add something to differentiate function calls from returns? Perhaps
```
>> A.adj(Array(32,))
>> return Array(128,)
```

Michael-T-McCann (changes requested):
Awesome! I added a few comments.

examples/scripts/trace_example.py
bwohlberg: Intentionally omitted from the notebooks since it's expected to primarily be of interest to developers.

examples/scripts/trace_example.py
bwohlberg: Done.

examples/scripts/trace_example.py
bwohlberg: How about this?
```
>> A.adj(Array(32,))
<< Array(128,)
```

Michael-T-McCann (approved):

PR #575, opened 2025-01-29
Docs fix

bwohlberg:
Docs fix and add utility function.

### DIFF
--- a/scico/linop/xray/astra.py
+++ b/scico/linop/xray/astra.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright (C) 2020-2024 by SCICO Developers
+# Copyright (C) 2020-2025 by SCICO Developers
 # All rights reserved. BSD 3-clause License.
 # This file is part of the SCICO package. Details of the copyright and
 # user license can be found in the 'LICENSE' file distributed with the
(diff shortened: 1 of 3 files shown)


scico/linop/xray/astra.py
```diff
@@ -441,8 +443,15 @@ def convert_to_scico_geometry(
...
-        proj_geom: ASTRA projection geometry object.
+        input_shape: Shape of the input array.
+        det_count: Number of detector elements. See the
+           `astra documentation <https://www.astra-toolbox.com/docs/geom3d.html#projection-geometries>`__
+           for more information.
+        det_spacing: Spacing between detector elements. See the
+           `astra documentation <https://www.astra-toolbox.com/docs/geom3d.html#projection-geometries>`__
+           for more information.
+        angles: Array of projection angles in radians.
+        vectors: Array of geometry specification vectors.
```
Michael-T-McCann, line 454: Slightly confusing that you can specify both of these, but (I think) only one ever gets used.

Michael-T-McCann (approved):

scico/linop/xray/astra.py
bwohlberg: I'll make an issue to note that this needs to be addressed.

Issue #576, opened 2025-02-04
Omissions in `xray.astra` docs


bwohlberg:
There are some minor omissions in the `xray.astra` docs:

Missing `Returns:` sections for `_project_coords`:
 https://github.com/lanl/scico/blob/a0e99bdd74a945bd7ebc0630c63c2207e3f3b5f3/scico/linop/xray/astra.py#L64-L76
and for `XRayTransform2D.fbp`:
https://github.com/lanl/scico/blob/a0e99bdd74a945bd7ebc0630c63c2207e3f3b5f3/scico/linop/xray/astra.py#L309-L320

Clarification needed on presence of `angles` and `vectors` at
https://github.com/lanl/scico/blob/a0e99bdd74a945bd7ebc0630c63c2207e3f3b5f3/scico/linop/xray/astra.py#L453-L454
and similar clarification could be provided in docs for `XRayTransform3D.__init__`.

Typo in line
https://github.com/lanl/scico/blob/a0e99bdd74a945bd7ebc0630c63c2207e3f3b5f3/scico/linop/xray/astra.py#L436

Use of "parallel3d_vec" format appears to be incompatible with `angles` option referred to above:
https://github.com/lanl/scico/blob/a0e99bdd74a945bd7ebc0630c63c2207e3f3b5f3/scico/linop/xray/astra.py#L439

It would also be worth adding some discussion of the meaning of "world coordinates" referred to in functions such as `project_world_coordinates`.

bwohlberg:
Resolved by #578.

PR #578, opened 2025-02-11
Improve X-ray docs

Michael-T-McCann:
Addresses #576. Is there a better place to put the text about world coordinates?

### DIFF
--- a/scico/linop/xray/astra.py
+++ b/scico/linop/xray/astra.py
@@ -74,7 +74,7 @@ def _project_coords(
         proj_geom: ASTRA projection geometry object.
 
     Returns:
-        (num_angles, ..., 3) array of detector coordinates corresponding
+        (num_angles, ..., 2) array of detector coordinates corresponding
         to projections of the points in `x_volume`.
 
     """
@@ -231,10 +231,10 @@ class XRayTransform2D(LinearOperator):
         self.num_dims = len(input_shape)
         if self.num_dims != 2:
             raise ValueError(
-                f"Only 2D projections are supported, but input_shape is {input_shape}."
+                f"Only 2D projections are supported, but 'input_shape' is {input_shape}."
             )
         if not isinstance(det_count, int):
-            raise ValueError("Expected det_count to be an int.")
+            raise ValueError("Expected 'det_count' to be an int.")
         output_shape: Shape = (len(angles), det_count)
 
         # Set up all the ASTRA config
@@ -256,7 +256,7 @@ class XRayTransform2D(LinearOperator):
                 self.vol_geom = astra.create_vol_geom(*input_shape, *volume_geometry)
             else:
                 raise ValueError(
-                    "volume_geometry must be a tuple of len 4."
+                    "Parameter 'volume_geometry' must be a tuple of len 4."
                     "Please see the astra documentation for details."
                 )
 
@@ -271,7 +271,7 @@ class XRayTransform2D(LinearOperator):
             dev0 = jax.devices()[0]
             self.device = dev0.platform
         else:
-            raise ValueError(f"Invalid device specified; got {device}.")
+            raise ValueError(f"Invalid 'device' specified; got {device}.")
 
         if self.device == "cpu":
             self.proj_id = astra.create_projector("line", self.proj_geom, self.vol_geom)
@@ -458,19 +458,19 @@ def convert_to_scico_geometry(
         det_spacing: Spacing between detector elements. See the
            `astra documentation <https://www.astra-toolbox.com/docs/geom3d.html#projection-geometries>`__
            for more information.
-        angles: Array of projection angles in radians, mutually
-            exclusive with `vectors`.
-        vectors: Array of ASTRA geometry specification vectors, mutually
-            exclusive with `angles`.
+        angles: Array of projection angles in radians. This parameter is
+            mutually exclusive with `vectors`.
+        vectors: Array of ASTRA geometry specification vectors. This
+            parameter is mutually exclusive with `angles`.
 
     Returns:
         (num_angles, 2, 4) array of homogeneous projection matrices.
 
     """
     if angles is not None and vectors is not None:
-        raise ValueError("`angles` and `vectors` are mutually exclusive.")
+        raise ValueError("Parameters 'angles' and 'vectors' are mutually exclusive.")
     if angles is None and vectors is None:
-        raise ValueError("One of `angles` and `vectors` must be provided.")
+        raise ValueError("Exactly one of parameters 'angles' and 'vectors' must be provided.")
     vol_geom, proj_geom = XRayTransform3D.create_astra_geometry(
         input_shape, det_count, det_spacing=det_spacing, angles=angles, vectors=vectors
     )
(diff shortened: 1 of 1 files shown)


scico/linop/xray/astra.py
```diff
@@ -450,13 +458,19 @@ def convert_to_scico_geometry(
...
+            exclusive with `vectors`.
+        vectors: Array of ASTRA geometry specification vectors, mutually
+            exclusive with `angles`.
 
     Returns:
         (num_angles, 2, 4) array of homogeneous projection matrices.
 
     """
+    if angles is not None and vectors is not None:
+        raise ValueError("`angles` and `vectors` are mutually exclusive.")
```
bwohlberg, line 471: Perhaps "... mutually exclusive parameters" for clarity?

scico/linop/xray/astra.py
```diff
@@ -577,8 +591,10 @@ def __init__(
             det_spacing: Spacing between detector elements. See the
                `astra documentation <https://www.astra-toolbox.com/docs/geom3d.html#projection-geometries>`__
                for more information.
-            angles: Array of projection angles in radians.
-            vectors: Array of geometry specification vectors.
+            angles: Array of projection angles in radians, mutually
+                exclusive with `vectors`.
```
bwohlberg, line 595: Perhaps "... radians. This parameter is mutually exclusive with ...", and likewise for the line below?

Also, before "Args:", "These options are ..." would be better as "These parameters are ..."

scico/linop/xray/astra.py
```diff
@@ -604,12 +620,15 @@ def __init__(
 
         if not isinstance(det_count, (list, tuple)) or len(det_count) != 2:
             raise ValueError("Expected det_count to be a tuple with 2 elements.")
+        if angles is not None and vectors is not None:
+            raise ValueError("`angles` and `vectors` are mutually exclusive.")
```
bwohlberg, line 624: Perhaps "Parameters angles and vectors ..."? Also, exceptions typically go to a text console, so no point in adding markup.

scico/linop/xray/astra.py
```diff
@@ -65,13 +65,18 @@ def _project_coords(
     x_volume: np.ndarray, vol_geom: VolumeGeometry, proj_geom: ProjectionGeometry
 ) -> np.ndarray:
     """
-    Transform volume (logical) coordinates into world coordinates based
+    Project volume (logical) coordinates into detector coordinates based
```
bwohlberg, line 68: Is this a change of terminology from "world coordinates" to "detector coordinates"? If so, corresponding changes are required elsewhere for consistency.

scico/linop/xray/astra.py
```diff
@@ -65,13 +65,18 @@ def _project_coords(
...
     on ASTRA geometry objects.
 
     Args:
         x_volume: (..., 3) vector(s) of volume (AKA logical) coordinates
```
bwohlberg, line 72: AKA is a bit jarring. Perhaps "i.e."?

scico/linop/xray/astra.py
```diff
@@ -95,7 +100,10 @@ def project_world_coordinates(
     """Project world coordinates along ray into the specified basis.
 
     Project world coordinates along `ray` into the basis described by `u`
-    and `v` with center `d`.
+    and `v` with center `d`. The term ""world"" emphasizes that the
+    function is intended to be used on 3D coordinates representing a
+    point in physical space, rather than a logical index into the volume
```
bwohlberg, line 105: "logical" implies an integer index into an array?

scico/linop/xray/astra.py
```diff
@@ -95,7 +100,10 @@ def project_world_coordinates(
     """Project world coordinates along ray into the specified basis.
 
     Project world coordinates along `ray` into the basis described by `u`
-    and `v` with center `d`.
+    and `v` with center `d`. The term ""world"" emphasizes that the
```
bwohlberg, line 103: Why double quotes on either side of "world"?

bwohlberg:
> ... Is there a better place to put the text about world coordinates?

The discussion *might* work better in the module docstring at the beginning of the file.

scico/linop/xray/astra.py
Michael-T-McCann: Understood about error messages going to the console, but I do think it's useful to have some marker that we are talking about a parameter name. E.g., numpy uses single quotes.

```
 sum() missing 1 required positional argument: 'a'
```

scico/linop/xray/astra.py
Michael-T-McCann: Not a change in terminology, the previous docstring was incorrect. 

Re volume vs logical: I hoped "logical" communicates that they are like integer indexes, but I'm removing it because I do not think it helps.

scico/linop/xray/astra.py
Michael-T-McCann: Remove as noted above.

scico/linop/xray/astra.py
Michael-T-McCann: Removed, I don't think "logical" helps explain anything.

scico/linop/xray/astra.py
Michael-T-McCann: I want that to print as 
```
The term "world" emphasizes...
```
using the quotes to indicate I'm talking about a word as word. I think the double double quotes escapes the double quotes to achieve that.

scico/linop/xray/astra.py
bwohlberg: Fair point. But for consistency we should make a note (perhaps a new issue) to also apply that principle elsewhere in the code.

scico/linop/xray/astra.py
bwohlberg: It looks as if "logical" is still present.

scico/linop/xray/astra.py
bwohlberg: I still see "AKA logical".

scico/linop/xray/astra.py
bwohlberg: It would also be good to mention somewhere what "world" coordinates mean.

scico/linop/xray/astra.py
bwohlberg: "logical" still present?

scico/linop/xray/astra.py
bwohlberg: I couldn't confirm this since the function in question isn't included in the rendered API docs, but it would be worth checking that this syntax usage is correct in case this function ever does get included in the docs.

bwohlberg (approved):

scico/linop/xray/astra.py
Michael-T-McCann: Issue #583 created.

scico/linop/xray/astra.py
Michael-T-McCann: Added to the module docstring after all.

Michael-T-McCann:
> > > ... Is there a better place to put the text about world coordinates?
> > 
> > 
> > The discussion _might_ work better in the module docstring at the beginning of the file.
> 
> No comment?

Added a discussion of this to the module docstring.

bwohlberg (approved):

PR #579, opened 2025-03-05
Add support for optimizer checkpointing

bwohlberg:
Add support for optimizer checkpointing. Also bump supported `jaxlib`/`jax` versions and address issues in docs.

### DIFF
--- a/scico/test/optimize/test_admm.py
+++ b/scico/test/optimize/test_admm.py
@@ -148,10 +148,31 @@ class TestReal:
         x = admm_.solve()
         assert (snp.linalg.norm(self.grdA(x) - self.grdb) / snp.linalg.norm(self.grdb)) < 1e-3
 
+    def test_admm_saveload(self):
+        maxiter = 5
+        ρ = 2e-1
+        A = linop.MatrixOperator(self.Amx)
+        f = loss.SquaredL2Loss(y=self.y, A=A, scale=self.𝛼 / 2.0)
+        g_list = [(self.λ / 2) * functional.SquaredL2Norm()]
+        C_list = [linop.MatrixOperator(self.Bmx)]
+        rho_list = [ρ]
+        admm0 = ADMM(
+            f=f,
+            g_list=g_list,
+            C_list=C_list,
+            rho_list=rho_list,
+            maxiter=maxiter,
+            itstat_options={"display": False},
+            x0=A.adj(self.y),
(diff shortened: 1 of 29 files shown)


scico/test/optimize/test_admm.py
```diff
@@ -145,6 +148,49 @@ def test_admm_generic(self):
...
+            rho_list=[ρ],
+            x0=A.adj(y),
+            maxiter=maxiter,
+            subproblem_solver=CircularConvolveSolver(),
+        )
+        admm0.solve()
+        with tempfile.TemporaryDirectory() as tmpdir:
+            path = os.path.join(tmpdir, "admm.npz")
+            admm0.save_state(path)
+            admm0.solve()
```
Michael-T-McCann, line 177: Why is `admm0.solve()` called twice?

scico/test/optimize/test_admm.py
```diff
@@ -145,6 +148,49 @@ def test_admm_generic(self):
         x = admm_.solve()
         assert (snp.linalg.norm(self.grdA(x) - self.grdb) / snp.linalg.norm(self.grdb)) < 1e-3
 
+    def test_admm_saveload(self):
```
Michael-T-McCann, line 151: Does this test represent the intended use case for state saving? It doesn't look like it to me, but I'm not sure. Could we instead do something like (1) create optimizer A (2) run A for e.g. 2 steps, save the state (3) create optimizer B, load the state from A, run 2 steps. (4) create C, run for 4 steps. (5) compare the result of B and C.

scico/test/optimize/test_admm.py
bwohlberg: The sequence is as follows:

1. Solve 5 iterations of `admm0`.
2. Save `admm0` state.
3. Solve another 5 iterations of `admm0`.
4. Initialize `admm1` with state of `admm0` after 5 iterations.
5. Solve another 5 iterations of `admm1`.
6. Compare results for `admm0` and `admm`.

The idea is that the additional 5 iterations after saving the state allow the solutions to diverge if any component of the state is not properly saved, so that the final test can just be on the solution rather than on all state components.

scico/test/optimize/test_admm.py
bwohlberg: It's not exactly replicating the intended use case, which is to be able to restart from a checkpoint. The approach you describe is quite similar to the current test - see details in my other response.

PR #596, opened 2025-08-01
Add cone-beam X-ray transform for symmetric volumes

bwohlberg:
Add cone-beam X-ray transform for symmetric volumes (a cone-beam variant of the Abel transform).


examples/scripts/ct_symcone_tv_padmm.py
```diff
@@ -0,0 +1,170 @@
...
+Create a ground truth image.
+"""
+N = 256  # image size
+x_gt = create_circular_phantom((N, N), [0.4 * N, 0.2 * N, 0.1 * N], [1, 0, 0.5])
+
+
+"""
+Set up the forward operator and create a test measurement.
+"""
+C = SymConeXRayTransform(x_gt.shape, 5e2 * N, 6e2 * N, num_slabs=4)
```
Michael-T-McCann, line 44: It would be more clear to use named arguments for the `obj_dist` and `det_dist`.

scico/linop/xray/_axitom/backprojection.py
```diff
@@ -0,0 +1,171 @@
...
+            )
+            - config.detector_vs[0]
+        ) / config.pixel_size_v
+    else:
+        raise ValueError("Invalid dimensions on the object coordinates")
+
+    return detector_cords_a, detector_cords_b
+
+
+def _fdk_axisym(projection_filtered, config, angles):
```
Michael-T-McCann, line 76: Maybe `_backproj_axisym` is a better name? The docstring is also misleading because it doesn't include filtering (as far as I can tell).

bwohlberg: True, but this name is from the original `axitom` code: addressing minor issues like this would be a slippery slope.

Michael-T-McCann: Ah, very reasonable.

Michael-T-McCann (approved):

Issue #592, opened 2025-07-15
Comment clarity in `scico/numpy/_blockarray`


bwohlberg:
The comments in matplotlib get backend make frequent reference to "BA"s and "DA"s, presumably referring to Block Array and Device Array respectively, e.g.

https://github.com/lanl/scico/blob/9017d7c7683ca5a5756b6d13f23620577002fa55/scico/numpy/_blockarray.py#L176

It would be helpful, though, to define these acronyms at first use, and perhaps also replace Device Array (DA) with e.g. Jax Array (JA), since device array is a bit more opaque, and seems no longer be frequently used terminology?

PR #597, opened 2025-08-22
Remove BA and DA abbreviations

Michael-T-McCann:
Addresses #592


bwohlberg (approved):
More extensive changes than the simple explanation of the abbreviations requested in #592, but the solution is even better.

PR #598, opened 2025-08-22
Resolve #583

### LINKED ISSUE #583: Use of variable names in error messages
Our use of variable names in error messages is inconsistent. In some places, we use single quotes around variables names,
```python
raise ValueError("Expected 'det_count' to be an int.")
```
and elsewhere we do not,
```python
raise ValueError("Initializer parameter l2_axis must be None for BlockArray input.")
```

I suggest generally following the former, which appears to match numpy style,
```bash
TypeError: sum() missing 1 required positional argument: 'a'
```
while also trying to use the word "argument" or "parameter" for additional clarity.

bwohlberg:
Resolve #583.

Some error messages could be further improved via much more substantial edits.

### DIFF
--- a/scico/flax/examples/examples.py
+++ b/scico/flax/examples/examples.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright (C) 2022-2024 by SCICO Developers
+# Copyright (C) 2022-2025 by SCICO Developers
 # All rights reserved. BSD 3-clause License.
 # This file is part of the SCICO package. Details of the copyright and
 # user license can be found in the 'LICENSE' file distributed with the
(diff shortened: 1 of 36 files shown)


Michael-T-McCann (commented):
If we are getting pedantic about parameters vs arguments, I do not think (though I am not certain) it is correct to say, e.g., "Argument `foo` has the wrong type." Rather, it should be "Argument passed to [parameter] `foo` has the wrong type," or "Parameter `foo` expected an argument of type X."

Refs that do not provide a definitive answer:
https://docs.python.org/3/glossary.html#term-parameter
https://docs.python.org/3/glossary.html#term-argument
https://docs.python.org/3/faq/programming.html#faq-argument-vs-parameter

scico/flax/examples/examples.py
```diff
@@ -682,7 +682,7 @@ def runtime_error_array(type: str, idstring: str, maxdiff: float):
            entries.
     """
     raise RuntimeError(
-        f"Requested value of parameter --{type}-- does not match value "
+        f"Requested value of parameter '{type}' does not match value "
```
crstngc, line 685: For consistency with the other edits, shouldn't this be: `f"Requested value of argument '{type}'"` ...?

crstngc (approved):

PR #603, opened 2025-09-26
Make `orbax-checkpoint` optional

bwohlberg:
Make `orbax-checkpoint` optional.


data
Michael-T-McCann, line 1: Are these updates to `/data/` expected? If so, no problem.

bwohlberg: No, thanks for catching that.

bwohlberg: Now there are legitimate changes, consisting of overlooked `scico-data` changes from earlier PRs.

PR #610, opened 2025-10-20
Update `jaxlib`/`jax` version

bwohlberg:
- Update `jaxlib`/`jax` version
- Fixes so that `astra` is an optional dependency.

### DIFF
--- a/scico/linop/xray/_util.py
+++ b/scico/linop/xray/_util.py
@@ -25,10 +25,8 @@ try:
 except ModuleNotFoundError as e:
     if e.name == "astra":
         have_astra = False
-        new_e = ModuleNotFoundError("Could not import astra; please install the ASTRA toolbox.")
     else:
-        new_e.name = "astra"
-        raise new_e from e
+        raise e
 import scipy.spatial.transform as sst
 
 


scico/linop/xray/_util.py
```diff
@@ -173,6 +183,8 @@ def image_alignment_rotation(
 
 
     """
+    if not have_astra:
+        raise RuntimeError("Package astra is required for use of this function.")
```
Michael-T-McCann: Consider simply `import astra` here without any of the other mechanism above.

scico/test/linop/xray/test_xray_util.py
```diff
@@ -35,6 +47,7 @@ def test_rotate_volume():
     np.testing.assert_allclose(vol.transpose((1, 2, 0)), vol_rot, rtol=1e-7)
 
 
+@pytest.mark.skipif(not have_astra, reason="astra not installed")
```
Michael-T-McCann: Alternative: https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-importorskip and remove the import logic at the top of the file.

scico/linop/xray/_util.py
bwohlberg: The goal is to make `astra` an optional dependency. This mechanism is in place because there's only one function in the module that uses `astra`.

Michael-T-McCann: Placing `import astra` in the body of that one function accomplishes the goal of making astra optional in a single line, whereas the current mechanism uses 10+ lines.

scico/test/linop/xray/test_xray_util.py
Michael-T-McCann: Just to make the benefit clear, `pytest.importorskip` would remove the need for the try except statement at the top of the file.

scico/linop/xray/_util.py
```diff
@@ -18,7 +18,17 @@
...
+    import scico.linop.xray.astra
+
+    have_astra = True
+except ModuleNotFoundError as e:
+    if e.name == "astra":
+        have_astra = False
+        new_e = ModuleNotFoundError("Could not import astra; please install the ASTRA toolbox.")
+    else:
+        new_e.name = "astra"
+        raise new_e from e
```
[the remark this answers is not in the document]
bwohlberg: This `else` would be reached if the import error were triggered by a module other than `astra`. The code was copied from elsewhere in scico, but indeed, it's not clear what the renaming is for. I shall remove it.

scico/test/linop/xray/test_xray_util.py
bwohlberg: The reason for keeping it as-is is largely for consistency with the practice elsewhere in scico.

scico/linop/xray/_util.py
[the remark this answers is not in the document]
bwohlberg: Fixed.

PR #623, opened 2025-12-18
Resolve issue #486

### LINKED ISSUE #486: Potential issues in `scico.solver`
The comment on this line in `scico.solver.minimize`

https://github.com/lanl/scico/blob/4eb21c103b5ec135f7a5a05f933cb0e58b7e64af/scico/solver.py#L216

is incorrect. This should be edited, and it should be confirmed that the functions does work for `BlockArray` input.

It's not clear whether the use of `nonlocal` and the "with side effects" is really necessary

https://github.com/lanl/scico/blob/4eb21c103b5ec135f7a5a05f933cb0e58b7e64af/scico/solver.py#L233-L250

It would be worth confirming this or simplifying the code.

bwohlberg:
Resolve issue #486.


Michael-T-McCann (changes requested):
I'd like to try cleaning this up further.

kmk142789 (dismissed):

Michael-T-McCann (approved):
I'm inclined to merge this because it improves the `BlockArray` handling. 

I think the use of `pure_callback` is superfluous because we don't intend to JIT code that calls `minimize`, but fixing that is separate, harder, and lower priority.

PR #640, opened 2026-04-09
Make snp.ravel fully flatten a BlockArray

### LINKED ISSUE #637: `BlockArray` does not have a function that flattens all blocks
A `scico.numpy.flatten` function should be added that is equivalent to the `flatten` method for `jax` arrays, and concatenates all flattened blocks for `BlockArray`s.

Michael-T-McCann:
See the new test for the proposed behavior of `snp.ravel` `BlockArray.ravel` and `BlockArray.flatten`. Closes #637.

### DIFF
--- a/scico/numpy/__init__.py
+++ b/scico/numpy/__init__.py
@@ -66,8 +66,7 @@ def ravel(ba: Union[Array | BlockArray]) -> Array:
     When called on an ``Array``, flattens the array.
 
     Args:
-        ba:
-           The :class:`BlockArray` to flatten.
+        ba: The :class:`BlockArray` to flatten.
 
     Returns:
         `ba` flattened into a single ``Array.``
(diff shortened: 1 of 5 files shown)


bwohlberg:
Some tests are failing, presumably where the `ravel` method and function are assumed to be equivalent.

scico/numpy/__init__.py
```diff
@@ -55,6 +59,25 @@
...
 
+
+def ravel(ba: Union[Array | BlockArray]) -> Array:
+    """Completely flatten a :class:`BlockArray` into a single ``Array``.
+
+    When called on an ``Array``, flattens the array.
+
+    Args:
+        ba:
+           The :class:`BlockArray` to flatten.
```
bwohlberg: Spurious linefeed: should be "ba: The ..."

bwohlberg (approved):

Issue #645, opened 2026-04-16
Arrays instantiated where `jax.eval_shape` should be used


bwohlberg:
In a number of locations in the scico source, the output shape and dtype of a function are determined by calling the function with a (potentially large) array of zeros as the argument:

https://github.com/lanl/scico/blob/5d0f9a486acefe54f27a1e2fa0ff36afb5a894e6/scico/operator/_operator.py#L152-L164

https://github.com/lanl/scico/blob/5d0f9a486acefe54f27a1e2fa0ff36afb5a894e6/scico/linop/_func.py#L125-L126

https://github.com/lanl/scico/blob/5d0f9a486acefe54f27a1e2fa0ff36afb5a894e6/scico/function.py#L71-L86

In each of these cases, the more efficient solution is to call 
```python
jax.eval_shape(fun, jax.ShapeDtypeStruct(input_shape, dtype=input_dtype))
```
to determine  the output shape and dtype. This is not currently possible, however, because `jax.eval_shape` does not support scico `BlockArray` shapes.

bwohlberg:
Based on the following test case, is appears that the solution is to use a list of `ShapeDtypeStruct` (one per block) for a `BlockArray` shaoe:
```python
import scico.numpy as snp
import jax
import numpy as np


def foo(x, y):
    return x[0]*y[0] + x[1]*y[1]


x = snp.ones([[2, 2], [2, 2]])
y = 2 * snp.ones([[1,], [1,]])


args = [list(jax.ShapeDtypeStruct(b_i.shape, b_i.dtype) for b_i in x),
        list(jax.ShapeDtypeStruct(b_i.shape, b_i.dtype) for b_i in y)]

jax.eval_shape(foo, *args)
```
Unfortunately it appears that this only works because the output is an array rather than a blockarray. If we change the evaluated function to
```python
def foo(x, y):
   return x * y
```
then the same approach fails. It's also worth noting that `jax.eval_shape` also fails when called as
```python
jax.eval_shape(foo, x, y)
```
i.e. with actual `BlockArray` arguments rather than lists of `ShapeDtypeStruct` to represent them.

PR #650, opened 2026-04-24
Refactor BlockArray again

Michael-T-McCann:
Currently a messy draft with remaining test failures, but useful for discussion.

The main idea is that we need an automatically mapping container that doesn't require its elements to be arrays. This allows code like
```python
eval_shape(
    foo,
    BlockArray(ShapeDtypeStruct(block.shape, block.dtype) for block in x),
    BlockArray(ShapeDtypeStruct(block.shape, block.dtype) for block in y),
)
```
This container is currently called `TransparentTuple`.

The change raises a number of questions:
* What to call the container (`MappingTuple`, `AutomappingTuple`, ...)
* How consistently should the container always map methods called on it? In particular, `==` is problematic. For block arrays, we usually want (broadcasting) elementwise comparison. But for shapes, we want a single bool out.
* Should BlockArray be a separate class from the container, or just what we call the container when it is full of arrays?

Relates to #645


Michael-T-McCann:
While there are good ideas here, it's not done and difficult to finish. We like having a more general mapping container (here called `TransparentTuple`). Ideally it could even allow arbitrary nesting. The complication comes around which operations shouldn't map (e.g., == should map for arrays but probably not for tuples) and how the API would work.

PR #651, opened 2026-04-28
Make BlockArray work with eval_shape

Michael-T-McCann:
The purpose of this branch is to discover how `eval_shape` breaks with `BlockArray` arguments. Currently one test in `test_blockarray.py` shows the mechanism working, but we expect it will fail if `eval_shape` is used more widely.

Relates to #645.


bwohlberg (approved):

bwohlberg:
@Michael-T-McCann: Since I made some substantial changes, you should also review before merging.

PR #653, opened 2026-05-06
Resolve #645

### LINKED ISSUE #645: Arrays instantiated where `jax.eval_shape` should be used

bwohlberg:
Resolve #645.


Michael-T-McCann (approved):
Cool that this works!

PR #660, opened 2026-06-02
Improve tangle phantom

bwohlberg:
Improve tangle phantom.

### DIFF
--- a/scico/examples.py
+++ b/scico/examples.py
@@ -538,9 +538,9 @@ def create_tangle_phantom(nx: int, ny: int, nz: int) -> np.ndarray:
         An array with shape (nz, ny, nx).
 
     """
-    xs = 1.0 * np.linspace(-1.0, 1.0, nx)
-    ys = 1.0 * np.linspace(-1.0, 1.0, ny)
-    zs = 1.0 * np.linspace(-1.0, 1.0, nz)
+    xs = 1.0 * np.linspace(-1.0, 1.0, nx, dtype=np.float32)
+    ys = 1.0 * np.linspace(-1.0, 1.0, ny, dtype=np.float32)
+    zs = 1.0 * np.linspace(-1.0, 1.0, nz, dtype=np.float32)
 
     # default ordering for meshgrid is `xy`, this makes inputs of length
     # M, N, P will create a mesh of N, M, P. Thus we want ys, zs and xs.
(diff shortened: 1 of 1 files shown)


scico/examples.py
```diff
@@ -560,7 +560,10 @@ def create_tangle_phantom(nx: int, ny: int, nz: int) -> np.ndarray:
         - 5.0 * zz * zz
         + 11.8
     ) * 0.2 + 0.5
-    return (values < 2.0).astype(float)
+    values[values <= 2.0] = 2.0 - values[values <= 2.0]
+    values[values > 2.0] = 0.0
+    values[values < 0.0] = 0.0
+    return values
```
Michael-T-McCann, line 581: What is the goal of these lines?

scico/examples.py
Michael-T-McCann, line 547: Also called the tanglecube, as described at https://mathworld.wolfram.com/Tanglecube.html with some changes to <I'm not quite sure. I think I applied scaling to change the range of values, but I don't know why.>

.. math::
     0.2 \times (x^4 - 5x^2 + y^4 - 5y^2 + z^4 - 5z^2 + 11.8) + 0.5

Michael-T-McCann (changes requested):
Let's take this chance to explain what we are doing a little bit.

scico/examples.py
bwohlberg: It would be worth including that reference in the docstring.

scico/examples.py
bwohlberg: The existing volume is binary, which makes it completely featureless when viewing slices. These changes introduce a range of values, but clip so that it still has finite extent.

scico/examples.py
Michael-T-McCann:
```suggestion
    """Construct a 3D phantom based on the tangle function:
    
    .. math::
    t(x, y, z) = 0.2 \times (x^4 - 5x^2 + y^4 - 5y^2 + z^4 - 5z^2 + 11.8) + 0.5
    
    Also called the tanglecube, this function is described at https://mathworld.wolfram.com/Tanglecube.html. To make a useful phantom, we compute
    
    .. math::
    p(x, y, z) = \begin{cases} 2 - t(x, y, z) & \text{if } t(x, y, z) \le 2 //
                       0 & \text{otherwise.}
                       \end{cases} 
                       
    """
    
```

Michael-T-McCann: I can't easily check how this looks in sphinx but it should be close

Michael-T-McCann (changes requested):

Issue #662, opened 2026-06-16
Initialization of `scico.linop.xray.XRayTransform3D` objects fails


bwohlberg:
Initialization of `scico.linop.xray.XRayTransform3D` objects fails with a `TracerIntegerConversionError`. Minimal example:
```python
import numpy as np
from scico.linop.xray.astra import angle_to_vector, convert_to_scico_geometry
from scico.linop.xray import XRayTransform3D

vol_shape = (64, 256, 256)
det_shape = (217, 358)
det_spacing = (1.0, 1.0)
angles = np.linspace(0, 2 * np.pi, 180, endpoint=False, dtype=np.float32)

# Matrices converted from astra vector geometry
vectors = angle_to_vector(det_spacing, angles)
matrices = convert_to_scico_geometry(input_shape=vol_shape, det_count=det_shape, vectors=vectors)
X = XRayTransform3D(vol_shape, matrices, det_shape)
# Initialization fails

# Matrices constructed from angles
rot_X = np.pi/2
rot_Y = angles
angles = np.stack(np.broadcast_arrays(rot_X, rot_Y), axis=-1)
matrices = XRayTransform3D.matrices_from_euler_angles(
    vol_shape, det_shape, "XY", angles, degrees=False
)
X = XRayTransform3D(vol_shape, matrices, det_shape)
# Initialization also fails
```

It's not clear why the tests pass or why `examples/scripts/ct_projector_comparison_3d.py` doesn't also fail.

PR #665, opened 2026-06-18
Improve `linop.xray.XRayTransform3D`

bwohlberg:
Improvements to `linop.xray.XRayTransform3D`, in particular:
- Forward and adjoint operations could previously not be jitted.
- Now ≈5x faster than astra rather than ≈10x slower. 

Notes:
- Improvements implemented with the assistance of Gemini
- This branch incorporates the fixes from #662.
- The approach in this branch includes removal of the manual batching (for memory use reduction) approach in the original implementation, relying instead on `jax` tools for this purpose, including using `batch_size` in `jax.lax.map`, and `jax.remat`.
- The original manual batching approach is retained in branch [brendt/xray-loops](/lanl/scico/tree/brendt/xray-loops). While the implementation in this branch is cleaner, that branch does seem to be ≈10% faster. Further tests (including quantification of both relative computation times and memory usage) and discussion would be appropriate before deciding which solution to adopt.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ SCICO Release Notes
 
 Version 0.0.8   (unreleased)
 ----------------------------
+
+• Substantial computational improvement in ``linop.xray.XRayTransform3D``,
+  which is now faster than ``linop.xray.astra.XRayTransform3D``.
 • Enable certain parameters of array creation functions to trigger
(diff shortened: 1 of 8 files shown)


bwohlberg:
Need to check that merge with main after merge of #669 did not break anything.

bwohlberg:
Now that sharding is supported in `main`, it would be worth making sure that the X-ray transform in this branch is compatible with sharding.

PR #670, opened 2026-06-30
Add support for sharded arrays

bwohlberg:
Add support for sharded arrays.


scico/linop/_circconv.py
```diff
@@ -148,7 +150,6 @@ def __init__(
             input_dtype=input_dtype,
             output_dtype=output_dtype,
             jit=jit,
-            **kwargs,
```
Michael-T-McCann, line 151: Most of our linops pass kwargs to `super().__init__`, but I don't recall why. In this case, there's no need: the only arguments not already supplied are `eval_fn` and `adj_fn`, neither of which make sense to supply here.

So, this seems fine to remove.

scico/linop/_circconv.py
Michael-T-McCann, line 88: I think this line should be removed given that kwargs is unused

scico/linop/_circconv3d.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def __init__(
+        self,
+        h: snp.Array,
+        input_shape: Shape,
+        input_dtype: DType = snp.float32,
+        h_is_dft: bool = False,
+        h_center: Optional[Union[snp.Array, np.ndarray, Sequence, float, int]] = None,
+        jit: bool = True,
+        **kwargs,
```
Michael-T-McCann: unused?

scico/numpy/util.py
```diff
@@ -265,6 +265,59 @@ def get_shape(in_shape, ind_expr):
...
+        Padded array and a tuple of slices that can be used to crop the
+        array back to its original shape.
+    """
+    if isinstance(axes, int) != isinstance(divisors, int):
+        raise ValueError("axes and divisors must be of the same type.")
+    if isinstance(axes, int) and isinstance(divisors, int):
+        axes = (axes,)
+        divisors = (divisors,)
+    assert isinstance(axes, (list, tuple)) and isinstance(divisors, (list, tuple))
+    pad_spec = [(0, _padding(n, x.shape, axes, divisors)) for n in range(x.ndim)]
```
Michael-T-McCann: Can we do this with less indirection (removing `_padding`)? Something like
```suggestion
    all_divisors = [divisors[axes.index(n)] if n in axes else x.shape[n] for n in range(x.ndim)]
    pad_spec = [(0, int(np.ceil(shape[n] / d) * d) - shape[n] for n, d in zip(range(x.ndim), all_divisors)]
```

scico/optimize/_pgm.py
```diff
@@ -111,7 +108,8 @@ def x_step(self, v: Union[Array, BlockArray], L: float) -> Union[Array, BlockArr
         return PGM._x_step(self.f, self.g, v, L)
 
     @staticmethod
-    @partial(jax.jit, static_argnums=(0, 1))
+    # Jit temporarily disabled due to issues with sharded arrays
+    # @partial(jax.jit, static_argnums=(0, 1))
```
Michael-T-McCann, line 112: Is there an issue we can reference here?

scico/test/numpy/test_numpy_util.py
```diff
@@ -155,11 +156,21 @@ def test_slice_length_other(length, slc):
...
 
 
+def test_pad_to_divisible():
+    shape = (2, 3, 4, 5)
+    x = np.zeros(shape, dtype=np.float32)
+    axes = (0, 2, 3)
+    divisors = (4, 5, 5)
+    x_pad, crop_spec = pad_to_divisible(x, axes, divisors)
+    assert x_pad.shape == (4, 3, 5, 5)
+    assert x_pad[crop_spec].shape == shape
```
Michael-T-McCann, line 171:
```suggestion
def test_pad_to_divisible():
    shape = (2, 3, 4, 19)
    x = np.zeros(shape, dtype=np.float32)
    axes = (0, 2, 3)
    divisors = (4, 5, 5)
    x_pad, crop_spec = pad_to_divisible(x, axes, divisors)
    assert x_pad.shape == (4, 3, 5, 20)
    assert x_pad[crop_spec].shape == shape
```

scico/util.py
```diff
@@ -119,10 +137,49 @@ def device_info(devid: int = 0) -> str:  # pragma: no cover
...
+    info = []
+    for dev in sorted(shard.device_set, key=lambda d: d.id):
+        info.append(_dev_info_device(dev))
+    return info
+
+
+def device_info(
+    dev: Union[int, Device, Sharding] = 0, string: bool = True
+) -> Union[str, List[str]]:  # pragma: no cover
+    """Get a string describing the specified device(s).
```
Michael-T-McCann, line 161: Not suggesting a change, buy FYI this may have been a good use case for `functools.singledispatch`

Michael-T-McCann (approved):

scico/linop/_circconv.py
bwohlberg: It's passed to the `super()` initializer.

scico/linop/_circconv.py
bwohlberg: Almost all of them do. I also don't recall why, but since this issue is so pervasive, I'd prefer to fix it globally (after suitable consideration) in a separate PR rather than make isolated fixes here.

scico/linop/_circconv3d.py
bwohlberg: Removed.

scico/numpy/util.py
bwohlberg: Changed.

scico/optimize/_pgm.py
bwohlberg: Unfortunately not. Thus far it has not been possible to isolate the problem enough to enable submitting a jax issue.

scico/util.py
bwohlberg: Good point.

PR #676, opened 2026-08-12
Denoiser performance improvements

bwohlberg:
Improve conditional UNet denoiser performance and reduce memory requirements over large batches.


bwohlberg:
Note that `mypy` check is failing due to typing errors in the destination branch, not due to changes in this branch.

PR #678, opened 2026-09-01
Add support functions for laminography problems

bwohlberg:
Add support functions for laminography problems.

Companion PR lanl/scico-data#64 should be merged before this one.


examples/scripts/cl_tv_apgm.py
```diff
@@ -0,0 +1,160 @@
...
+
+  $$\mathrm{argmin}_{\mathbf{x}} \; (1/2) \| \mathbf{y} - A \mathbf{x}
+  \|_2^2 + \lambda \| C \mathbf{x} \|_{2,1} \;,$$
+
+where $A$ is the X-ray transform (the CT forward projection operator),
+$\mathbf{y}$ is the sinogram, $C$ is a 2D finite difference operator, and
+$\mathbf{x}$ is the reconstructed image. This example uses the CT
+projector integrated into scico, while the companion
+[example script](ct_astra_tv_admm.rst) uses the projector provided by
+the astra package.
```
Michael-T-McCann: This intro should mention computed laminography and probably shouldn't call out ct_astra_tv_admm.rst as its companion example.

scico/linop/xray/_cl.py
```diff
@@ -0,0 +1,124 @@
...
+           :math:`\times` Ncol pixels.
+        alpha: Laminography tilt angle (see angle :math:`\alpha` in Fig.
+           3(a) in :cite:`aarle-2016-fast`) in radians.
+        X: :class:`.xray.XRayTransform3D` or :class:`.astra.XRayTransform3D`
+           linear operator corresponding to the imaging geometry.
+
+    Returns:
+        FBP reconstruction.
+    """
+    yf = _filter_projection(y, alpha)
```
Michael-T-McCann, line 124: This will be wrong for the scico projector because the it assumes `(Nrows, Nview, Ncol)` order.

scico/linop/xray/_cl.py
```diff
@@ -0,0 +1,124 @@
...
+    # scico projector order is (Nview, Nrow, Ncol) while astra order
+    # is (Nrows, Nview, Ncol).
+    n_proj = y.shape[0] if isinstance(X, scicoXRayTransform3D) else y.shape[1]
```
Michael-T-McCann, line 127: Maybe being too clever. Require a specific dimension ordering, or have it passed as an argument.

scico/test/linop/xray/test_cl.py
```diff
@@ -0,0 +1,75 @@
...
+    vol = np.zeros((9, 33, 33), dtype=np.float32)
+    vol[4, 16, 16] = 1
+    det_shape = (24, 48)
+    matrices = convert_to_scico_geometry(
+        input_shape=vol.shape, det_count=det_shape, vectors=vectors
+    )
+    X = scicoXRayTransform3D(vol.shape, matrices, det_shape)
+    y = X @ vol
+    xfbp = cl_fbp(y, alpha, X)
+    assert xfbp[4, 16, 16] == xfbp.max()
```
Michael-T-McCann, line 56: I'm surprised this is passing, maybe I'm wrong about cl_fbp breaking for scico's x-ray transform. Or maybe xfbp is all zeros or similar.

scico/examples.py
```diff
@@ -619,6 +616,139 @@ def create_block_phantom(out_shape: Shape) -> np.ndarray:
...
+
+
+def create_laminar_phantom() -> np.ndarray:
+    """Construct a :math:`64 \times 256 \times 256` laminar phantom`.
+
+    Construct a :math:`64 \times 256 \times 256` laminar phantom`.
+
+    Returns:
+        Laminar phantom.
+    """
```
Michael-T-McCann, line 670: What does this phantom look like? It's a lot of lines of somewhat opaque code (including the `_extract_blocks` function). Maybe there is a simpler way to generate it.

Michael-T-McCann (changes requested):

scico/linop/xray/_cl.py
bwohlberg: Convolution is along 3rd axis, which is Ncol for both projectors. Comments added to clarify this point.

scico/linop/xray/_cl.py
bwohlberg: There are only two options. The required ordering is now documented.

scico/test/linop/xray/test_cl.py
bwohlberg: See comments above confirming that `cl_fbp` is not broken.

examples/scripts/cl_tv_apgm.py
bwohlberg: Fixed.

Michael-T-McCann (approved):

PR #679, opened 2026-09-21
Bump max `jax` version and fix tests

bwohlberg:
* Bump maximum `jax` version
* Resolve undefined class when `astra` not installed
* Fix `astra` presence tests in some test modules.


scico/linop/xray/_cl.py
```diff
@@ -20,7 +20,7 @@
 try:  # scico.astra cannot be imported if astra is not installed
     from .astra import XRayTransform3D as astraXRayTransform3D
 except ModuleNotFoundError:
-    pass
+    astraXRayTransform3D = scicoXRayTransform3D  # type: ignore
```
Michael-T-McCann: I find this confusing. I think it is happening just to support a type hint later (line 100). How about changing that type hint to `LinearOperator` and removing this import.

CHANGES.rst
```diff
@@ -12,7 +12,7 @@ Version 0.0.8   (unreleased)
 • New functional ``functional.BoxIndicator``.
 • New functions supporting computed laminography reconstructions.
 • Module ``scico.plot`` deprecated.
-• Support ``jaxlib`` and ``jax`` versions 0.5.0 to 0.11.1.
+• Support ``jaxlib`` and ``jax`` versions 0.5.0 to 0.11.2.
```
Michael-T-McCann: Does CI check the lower bound jaxlib and jax versions? I was curious and ran a local test; I can confirm they still work.

Michael-T-McCann (approved):

scico/linop/xray/_cl.py
bwohlberg: But it doesn't work for arbitrary LinearOperator. How about adding an explanatory comment instead?

CHANGES.rst
bwohlberg: No, it doesn't. I guess we should check periodically.

